PR #5, opened 2014-06-05
Add Constraints dictionary to Terminology.

anssiko:


anssiko:
This is an editorial fix, so I'll merge without review.

PR #6, opened 2014-06-06
Add an Introduction section; clarify 'depth stream' and 'depth track' in the Terminology section.

anssiko:


index.html
```diff
@@ -118,13 +128,12 @@
         type is defined in [[!typedarray]].
       </p>
       <p>
-        A <dfn>depth stream</dfn> is an abstraction of a data stream that
-        represents the distance between the camera and the subject for each
-        point of the image.
+        A <dfn>depth stream</dfn> is a <a><code>MediaStream</code></a> object
+        that contains one of more <a>depth track</a>s.
```
robman: I think on line 132 "one of more" should be "one or more".

PR #14, opened 2014-08-21
First stab at DepthData interface and CanvasRenderingContext2D extensions.

anssiko:


huningxin:
@anssiko , thanks for the pull!

index.html
```diff
@@ -232,20 +236,104 @@
...
+            -
+          </dd>
+          <dt>
+            readonly attribute unsigned long height
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute ArrayBufferView data
```
huningxin: Can we use Uint16Array here?

index.html
```diff
@@ -232,20 +236,104 @@
...
+          </dd>
+        </dl>
+        <p>
+          New <code><a>DepthData</a></code> objects MUST be initialised so that
+          their <code>width</code> attribute is set to the number of entries
+          per row in the <a>depth data</a>, their <code>height</code> attribute
+          is set to the number of rows in the <a>depth data</a>, and their
+          <code>data</code> attribute, except where an existing array is
+          provided, is initialised to a new <code>ArrayBufferView</code>
+          object.
```
huningxin: Maybe, we need to say something about the layout of data. Say data[x \* width + y] represents the depth value at position (x, y).

index.html
```diff
@@ -232,20 +236,104 @@
...
-            separate attribute to the ImageData for the Uint16Array type
-            without the need to patch the Canvas spec.
+            (Please note, that we may need to file bugs against [[!2DCONTEXT2]]
+            to add extension points this extension specification can hook into
+            to avoid monkey patching.)
           </p>
         </div>
+        <dl class="idl" title="[Constructor(unsigned long sw, unsigned long sh),
+ Constructor(ArrayBufferView data, unsigned long sw, optional unsigned long sh),
+ Exposed=Window,Worker] interface DepthData">
```
huningxin, line 272: Do we consider to add some meta data, such as `unit`, in the DepthData interface? As Benjamin mentioned.

index.html
anssiko: I changed ArrayBufferView into Uint16Array type.

I was thinking that for future-proofing (e.g. to allow Uint32Array usage in the future, as suggested by Benjamin at http://lists.w3.org/Archives/Public/public-media-capture/2014Aug/0102.html) we'd use ArrayBufferView similarly to e.g. WebSocket's send() does. But I think it makes the spec simpler if we use the Uint16Array type instead.

We can overload the constructor in future revisions of the spec, if the need arises.

index.html
anssiko: I added prose to the spec. HTML Canvas 2D Context spec defines this for RGB, so I adapted that prose to fit the data model used for depth data.

index.html
anssiko, line 272: I added CameraParameters interface that encapsulates this metadata. I added an issue to the spec too outlining some open issues.

anssiko:
@huningxin Please review the updated PR. If no major issues, I'll merge and we can ping the group for review.

Thanks for your review!

huningxin:
LGTM. Thanks so much!

PR #15, opened 2014-09-04
Update getusermedia.html ED references.

anssiko:


anssiko:
Since this is an editorial change, I'll merge without review.

PR #19, opened 2014-09-22
Add releases/FPWD snapshot.

anssiko:


anssiko:
This is the First Public Working Draft snapshot proposal (.src.html is the ReSpec source, .html the HTML snapshot) to be used for publication, assuming a publication date 02 October 2014.
- Pubrules (http://www.w3.org/2005/07/pubrules) - pass
- Link Checker (http://validator.w3.org/checklink) - pass, except some broken fragments that are expected

PR #24, opened 2014-10-08
Add 'Participate' header with pointers to ML and GH.

anssiko:


anssiko:
@huningxin @robman This is to make it easier for people outside the TF to participate too.

huningxin:
LGTM. Thanks for this PR.

PR #33, opened 2014-11-14
setting up Travis CI

dontcallmedom:
See http://lists.w3.org/Archives/Public/spec-prod/2014OctDec/0067.html and http://lists.w3.org/Archives/Public/spec-prod/2014OctDec/0076.html for context


huningxin:
LGTM. Thanks.

PR #35, opened 2014-11-25
Add WebGLRenderingContext interface section.

huningxin:


huningxin:
@anssiko @robman , please take a look. Thanks!

huningxin:
@anssiko @robman , any comments?

anssiko:
LGTM. We may want to add a note noting this section is informative, since the WebGLRenderingContext is defined in a Khronos spec, and thus this spec cannot impose normative requirements upon interfaces defined in that spec.

Created a new issue for that https://github.com/w3c/mediacapture-depth/issues/46

PR #54, opened 2015-01-23
Minor update to the 2d canvas example.

robman:


anssiko:
@robman LGTM. Thanks! Does this example pass JSLint now?

PR #58, opened 2015-01-23
An update to CameraIntrinsics to align with mediacapture-main.

anssiko:
@robman @ningxin I made some updates, please review. Updated the IDL and prose since dictionaries are a bit different in terms of WebIDL from interfaces. This is the typical boilerplate that is used for dictionaries.

### DIFF
--- a/index.html
+++ b/index.html
@@ -362,50 +362,52 @@
       </section>
       <section>
         <h2>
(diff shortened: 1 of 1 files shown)


anssiko:
Closed via https://github.com/w3c/mediacapture-depth/pull/59

PR #62, opened 2015-01-23
Add 'Changes since FPWD' to SoTD section.

anssiko:
@ningxin @robman Here's the changes since based on your feedback. I'll merge this right away without review -- we can update it in subsequent commits if needed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -84,6 +84,54 @@
     </section>
     <section id='sotd'>
       <p>
+        The following changes were made since the
+        <a href="http://www.w3.org/TR/2014/WD-mediacapture-depth-20141007/">
+        W3C First Public Working Draft 07 October 2014</a>:
+      </p>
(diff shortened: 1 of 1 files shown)


huningxin:
LGTM. Thanks for the PR.

One thing we might need to do is to notify WebGL community that we don't need https://www.khronos.org/registry/webgl/extensions/proposals/WEBGL_texture_from_depth_video/ and probably send a PR WebGL repo to remove it. 

@anssiko @robman , WDYT?

anssiko:
@huningxin Let's do that once we've released WD1. Rather than remove, I'd like to include a visible note stating the extensions is not longer considered active work. That is the practice that W3C prefers, I'd guess Khronos too.

huningxin:
Understood and thanks for the explanation.

PR #63, opened 2015-01-26
Editorial: Clean up with HTML Tidy for HTML5

anssiko:


anssiko:
Merging without review. This is a clean up of the markup using HTML Tidy for HTML5.

PR #64, opened 2015-01-26
Editorial changes

anssiko:


anssiko:
Editorial, thus merging without review.

Issue #67, opened 2015-02-07
Add depth context of HTML canvas element


huningxin:

ds-hwang:
could you point out mailing list or discussion related to this issue?

anssiko:
I think the proposal can be attributed to Rik: https://lists.w3.org/Archives/Public/public-canvas-api/2014OctDec/0034.html

huningxin:
As we made consensus on this one, can i close it?

ds-hwang:
+1

anssiko:
Please feel free to close this, I closed the related PR https://github.com/w3c/mediacapture-depth/pull/68

Issue #66, opened 2015-02-07
Add DepthMap interface


huningxin:
```
interface DepthMap {
    readonly attribute unsigned long width;
    readonly attribute unsigned long height;
    readonly attribute string type;
    readonly attribute string format;
    readonly attribute string units;
    readonly attribute float near;
    readonly attribute float far;
    readonly attribute Uint16Array dexels;
    readonly attribute Uint8ClampedArray? pixels;
};
```

huningxin:
Reference: https://developers.google.com/depthmap-metadata/reference

anssiko:
Closed via https://github.com/w3c/mediacapture-depth/pull/77

PR #68, opened 2015-02-10
First stab at the depth context.

anssiko:


anssiko:
@huningxin @robman I wrote down some of the initial ideas discussed. Please review. We should update this PR based on your feedback, if this looks like something we want to pursue further.

HTML preview at: https://rawgit.com/anssiko/mediacapture-depth/depthcontext/index.html#the-depth-context

anssiko:
Work in progress to address https://github.com/w3c/mediacapture-depth/issues/66 and https://github.com/w3c/mediacapture-depth/issues/67

index.html
```diff
@@ -498,6 +503,133 @@
...
+        </p>
+      </div>
+      <section>
+        <h3>
+          <code>DepthContext</code> interface
+        </h3>
+        <dl class="idl" title=
+        "[Constructor(optional unsigned long width, unsigned long height), Exposed=Window,Worker] interface DepthContext">
+        <dt>
+            DepthData getDepthData(double sx, double sy, double sw, double sh)
```
huningxin: The name `depthData` and `depthMap` are not consistent.

anssiko: @huningxin Good catch, fixed with https://github.com/anssiko/mediacapture-depth/commit/2b26c61ab355e00fa51eb0c39ac01c43e2d731d7

ds-hwang:
I don't understand why depth context should be introduced because of two reasons.
1. which value is added over CanvasImageSource? http://w3c.github.io/mediacapture-depth/#canvasimagesource-typedef
2. The interface looks inherently software path, which means GPU video frame should read back to CPU for this API.

IMO, depth camera's ultimate use case would be under WebCL by expending WEBCL_html_video Extension https://www.khronos.org/registry/webcl/extensions/WEBCL_html_video/
However, major browsers don't have plan to support WebCL currently. Until that time, IMO most depth camera user or javascript library should use WebGL, instead of both 2D canvas and this depth context.
What I mean is that CanvasImageSource is enough to support javascript software fallback.

anssiko:
@ds-hwang Thanks for the review. This PR is an experiment to see how a depth context might look like (along the lines of the [original proposal by Rik](https://lists.w3.org/Archives/Public/public-canvas-api/2014OctDec/0034.html)). After hearing your feedback and revisiting the related discussion on public-canvas-api@w3.org mailing list, e.g. [Justin's feedback](https://lists.w3.org/Archives/Public/public-canvas-api/2014OctDec/0048.html), I'm coming to a conclusion we should drop this depth context idea and perhaps explore the `getDepthTracks().getPixels()` idea coined by @robman instead (in the same thread above) and also document known issues with it.

@huningxin I think we can address https://github.com/w3c/mediacapture-depth/issues/66 by exposing relevant `DepthMap` metadata through `Settings` and/or on the depth stream (i.e. `MediaStream` that contains one or more depth tracks). Those are probably the only bits that should survive from this PR. We could also close https://github.com/w3c/mediacapture-depth/issues/67 if we drop this PR. WDYT? 

@robman please let me know if you'd like me to help you draft the `getDepthTracks().getPixels()` idea in another topic branch.

@ds-hwang For the background, see the initial [use cases and requirements](https://www.w3.org/wiki/Media_Capture_Depth_Stream_Extension) for the API. Note the examples in the wiki have not been updated to match the latest API, while the requirements should still hold (if that is not the case, we should revise them :-))

ds-hwang:
thx for pointing out previous discussion. 
[Mediastream Image Capture](http://w3c.github.io/mediacapture-image/index.html#bib-CANVAS-2D) 's `grabFrame()` API looks nice solution to cover both 2D canvas extension and depth context which is similar to getDepthTracks().getPixels()

huningxin:
> @ds-hwang Thanks for the review. This PR is an experiment to see how a depth context might look like (along the lines of the original proposal by Rik). After hearing your feedback and revisiting the related discussion on public-canvas-api@w3.org mailing list, e.g. Justin's feedback, I'm coming to a conclusion we should drop this depth context idea and perhaps explore the getDepthTracks().getPixels() idea coined by @robman instead (in the same thread above) and also document known issues with it.

@ds-hwang , Thanks! +1 to drop the depth context idea.

> @huningxin I think we can address #66 by exposing relevant DepthMap metadata through Settings and/or on the depth stream (i.e. MediaStream that contains one or more depth tracks). Those are probably the only bits that should survive from this PR. We could also close #67 if we drop this PR. WDYT?

The main idea of #66 is to expose depth attributes and depth value (in a more natural way). 

For depth attributes, I like the idea to expose them in MST setting level, as @robman mentioned. Say  

```
partial dictionary Settings {
    readonly attribute string type;
    readonly attribute string format;
    readonly attribute string units;
    readonly attribute float near;
    readonly attribute float far;
};
```

For depth value, we still need to figure out where to put the depth data (Uint16Array). 

One option would be extending ImageData interface with dexels. 

```
partial interface ImageData {
    readonly attribute Uint16Array? dexels;
};
```

With this extension, we can still leverage MS->Video->Canvas post-processing pipeline to handle three use cases:
1. 1 Depth Track
2.  1 Depth Track and 1 Video Track
3.  1 Video Track

I am splitting #66 to #72 and #73.

@anssiko , I am OK to close #67. 

> @robman please let me know if you'd like me to help you draft the getDepthTracks().getPixels() idea in another topic branch.

Interesting to head this direction. May I know how `getDepthTracks().getPixels()` solves "1 Depth Track and 1 Video Track" case?

ds-hwang:
`mediastream.getDepthTracks()[0].getPixels()` should return depth ImageData. If `getDepthTracks()[0].getPixels()` is implemented, `mediastream.getVideoTracks()[0].getPixels()` should be implemented in the same sense. In addition, this API looks duplicated to [mediastream image capture spec](http://w3c.github.io/mediacapture-image/index.html). I don't know which is better. IMO async api (i.e. mediastream image capture) might be overkill because implementation doesn't require waiting anything.

IMO ImageData should extend format instead of dexels

```
partial interface ImageData {
    readonly attribute int format;
};
```

`format` might consist of RGBA, RGB, DEPTH16

anssiko:
Closing this PR since we seem to be headed toward extending (or reusing) Mediastream Image Capture https://github.com/w3c/mediacapture-depth/issues/76

PR #70, opened 2015-02-23
small improvement of 2D canvas example

ds-hwang:


ds-hwang:
@anssiko @huningxin could you review?

anssiko:
@ds-hwang LGTM. Thanks for fixing the bug in the example and modernising it by means of using rAF.

ds-hwang:
thx for quick reviewing :)

huningxin:
@ds-hwang , LGTM! Thanks for the fixing!

PR #74, opened 2015-02-25
Fix #69: Define VideoTrack.kind

### LINKED ISSUE #69: Making Track definition more explicit and an extension of MCAS's definition
The relationship between MediaStreamTracks and HTMLVideoElement VideoTracks and AudioTracks wasn't 100% clear. The MCAS spec seems to have cleared this up quite a bit now.

  For each MediaStreamTrack in the MediaStream , including those that
  are added after the User Agent enters the media element load
  algorithm, the User Agent MUST create a corresponding AudioTrack or
  VideoTrack as defined in [HTML5].

http://w3c.github.io/mediacapture-main/#loading-and-playing-a-mediastream-in-a-media-element 

We should extend this specifically to support Depth.

anssiko:
@huningxin @robman Please review.

Bug: https://github.com/w3c/mediacapture-depth/issues/69

### DIFF
--- a/index.html
+++ b/index.html
@@ -350,10 +350,12 @@
         </p>
         <p>
           For each <a><code>MediaStreamTrack</code></a> representing a <a>depth
-          track</a> in the <code><a>MediaStream</a></code>, the <a>user
-          agent</a> MUST create a corresponding <code><a href=
+          track</a> in the <code><a>MediaStream</a></code> assigned directly to
+          a media element, the <a>user agent</a> MUST create a corresponding
+          <code><a href=
           "http://www.w3.org/TR/html5/embedded-content-0.html#videotrack">VideoTrack</a></code>
-          as defined in [[HTML5]].
+          as defined in [[HTML5]] whose <code>kind</code> attribute MUST, on
+          getting, return the string "<code>depth</code>".
         </p>
       </section>
       <section>


huningxin:
If we do this, we need to extend http://www.w3.org/TR/html5/embedded-content-0.html#dom-videotrack-kind to support 'depth'. However, @robman suggests we use "MST kind casting" to cast MST.kind='depth' to HTMLVideoElement.getVideoTracks()[0].kind='main'.

huningxin:
IMO, I think supporting 'depth' in VideoTrack.kind is natural, like 'alternative' kind which from a camera of a different angle. @robman , could you please elaborate what changes need to be made in html5 video for 'depth' VideoTrack.kind?

robman:
The problem is that MST.kind and VideoTrack.kind are different logical types.

MST.kind can be:
- video
- audio
- depth

VideoTrack.kind can be:
- main
- alternative
- etc. http://www.w3.org/TR/html5/embedded-content-0.html#dom-TrackList-getKind-categories

I take your point that "alternative" could work as the VideoTrack.kind for an MST.kind=depth - that's probably better than "main" in a <video> with both a video and depth track. But "depth" doesn't seem like the same logical type as "main" or "alternative" to me.

BTW: I think if we resolve fully how the calibrated video streams and ImageData objects work then this question will probably also be resolved 8)

Also, if you look at the extension in 6.2 of MCAS then it's clear they're mapping from MST.kind (e.g. video) to VideoTrack.kind (e.g. main). Not adding new peers to "main".

And my last point is one I made before 8) If we did this then I think we'd also need to pepper DepthTrack extensions all through the html5 spec where VideoTrack is defined. I agree this would probably be the "correct" thing to do but it seems like a lot of work that would probably be contentious.

Thoughts?

anssiko:
It sounds like we should step back, and discuss what is the use case for `VideoTrack.kind` "depth". I thought this would be a trivial fix to address https://github.com/w3c/mediacapture-depth/issues/69 but based on your feedback that seems not to be the case. I prefer not to land this without understanding the use case first.

If this is just a convenience feature that can be worked around, I'd rather not introduce a new "depth" category for `VideoTrack.kind` to minimize monkey patching needed for HTML5.

I filed https://github.com/w3c/mediacapture-depth/issues/75 in an attempt to clarify the expectations re synchronization and calibration across MediaStreams.

Issue #76, opened 2015-03-04
Extend Mediastream Image Capture to support depth track


huningxin:
The proposal is:

```
[Constructor(MediaStream stream)]
interface ImageCapture {
    readonly    attribute PhotoCapabilities photoCapabilities;
    readonly    attribute MediaStream  videoStream;
    readonly    attribute MediaStream       previewStream;
    Promise setOptions (PhotoSettings? photoSettings);
    Promise takePhoto ();
    Promise grabFrame ();
};

partial interface FrameGrabEvent {
    readonly attribute DepthMap depthMap;
};
```
1. Modify the ImageCapture to take MediaStream object which may carry both video track and depth track.
2. Extend the FrameGrabEvent to include DepthMap object which is filled based on depth track when grabbing. The DepthMap interface is defined in https://github.com/w3c/mediacapture-depth/issues/66.

The sample code sketch looks like:

```
<html>
<body>
<button onclick="stopFunction()">Stop frame grab</button>
<script>
var canvas = document.getElementById('frame');
navigator.getUserMedia({video: true, depth: true}, gotMedia, failedToGetMedia);

function gotMedia(mediastream) {
   // mediastream carries both video track and depth track
   var captureDevice = new ImageCapture(mediastream);
   var frameVar;
   if (captureDevice) {
       frameVar = setInterval(captureDevice.grabFrame().then(processFrame()), 1000);
   }
}

function processFrame(e) {
  var imgData = e.imageData;
  var depthMap = e.depthMap;
  ASSERT(imgData.width == depthMap.width);
  ASSERT(imgData.height == depthMap.height);
  var w = imgData.width, h = imgData.height;
  var pixels = imgData.data, dexels = depthMap.data;
  for (var x = 0; x < w ; ++x) {
    for (var y = 0; y < h; ++y) {
      var i = x + y * w;
      var r = pixels[i*4];
      var g = pixels[i*4 + 1];
      var b = pixels[i*4 + 2];
      var depth = dexels[i]; // Uint16 depth value
    }
  }
}

function stopFunction(e) {
  clearInterval(frameVar);
} 
</script>
</body>
</html>   
```

ds-hwang:
looks nice!

huningxin:
@robman @anssiko , please take a look. Thanks!

anssiko:
Thanks @huningxin for the proposal!

[...]

Issue #73, opened 2015-02-25
Extend ImageData to support depth data


huningxin:
Reference is: https://developers.google.com/depthmap-metadata/reference

The proposal is:

```
partial interface ImageData {
    readonly attribute Uint16Array? depth_data;
};
```

Sample code sketch is:

```
var canvasContext = document.createElement("canvas").getContext("2d");
var fps = 60;

navigator.mediaDevices.getUserMedia({
  video: true,
  depth: true,
}).then(function (stream) {
  // wire the stream into a <video> element for playback
  var video = document.querySelector('#video');
  video.srcObject = stream;  // stream carries both video and depth tracks
  video.play();

  video.onloadedmetadata = function () {
    function doSomething() {
      canvasContext.drawImage(video); // propagate both pixels and dexels.
      var imageData = canvasContext.getImageData(0, 0, w, h);
      var pixels = imageData.data;
      var dexels = imageData.depth_data;

      for (var x = 0; x < w ; ++x) {
        for (var y = 0; y < h; ++y) {
          var i = x + y * w;
          var r = pixels[i*4];
          var g = pixels[i*4 + 1];
          var b = pixels[i*4 + 2];
          var depth = dexels[i];
        }
      }
      window.requestAnimationFrame(doSomething);
    }
    window.requestAnimationFrame(doSomething);
  };
}).catch(function (reason) {
  // handle gUM error here
});

```

ds-hwang:
The example requires huge hack in canvas, because getImageData() intend to read back from framebuffer and the framebuffer don't have depth component in general.

```
      canvasContext.drawImage(video); // propagate both pixels and dexels.
      var imageData = canvasContext.getImageData(0, 0, w, h);
      var pixels = imageData.data;
      var dexels = imageData.depth_data;
```

Let me change the proposal a bit using existing [mediacapture image spec](http://w3c.github.io/mediacapture-image/index.html#repeated-grabbing-of-a-frame)

```
partial interface ImageData {
    readonly attribute int format; // RGBA or DEPTH16
};
```

The example is very similar to the example in mediacapture image spec

```
<html>
<body>
<button onclick="stopFunction()">Stop frame grab</button>
<script>
var canvas = document.getElementById('frame');
navigator.getUserMedia({video: true, depth: true}, gotMedia, failedToGetMedia);

function gotMedia(mediastream) {
   //Extract video track.  
   var videoDevice = mediastream.getVideoTracks()[0];
[...]

Issue #72, opened 2015-02-25
Add more depth relevant attributes into Settings interface


huningxin:
Reference is: https://developers.google.com/depthmap-metadata/reference

Proposal is:

```
partial dictionary Settings {
    readonly attribute string type;
    readonly attribute string format;
    readonly attribute string units;
    readonly attribute float  near;
    readonly attribute float  far;
};
```

anssiko:
It seems we should expose `Settings` via the `ImageCapture` interface and reuse the accessors it provides.

I suggested changes to the accessors of the mediacapture-image, see: https://lists.w3.org/Archives/Public/public-media-capture/2015Mar/0048.html We should update http://w3c.github.io/mediacapture-depth/#settings-dictionary accordingly when the suggested changes land to the mediacapture-image.

anssiko:
Closed via https://github.com/w3c/mediacapture-depth/pull/77

PR #77, opened 2015-04-22
Proposal for FrameGrabber, FrameData, DepthMap

anssiko:
This PR contains an early proposal with contributions from @huningxin @ds-hwang @robman and @anssiko. IDL is in place to allow people to review the API shape, but the prose around the new interfaces is still missing to allow us more easily iterate on the API design based on wider feedback.

HTML preview: https://rawgit.com/anssiko/mediacapture-depth/framegrabber/index.html

New interfaces added in this PR:
- `FrameGrabber`
- `FrameData`
- `DepthData`

The following interfaces and definitions were obsoleted by the new ones, and were removed:
- `CanvasImageSource typedef`
- `ImageData interface` (or to be precise, extensions to it)

Examples:
- Added a new example 'Capture individual depth and RGB frames'; removed the obsoleted '2D Canvas Context based post-processing' example.

Editorial:
- Removed the changes since the last publication from the Status of This Document section. Not needed for an Editor's Draft.

### DIFF
--- a/index.html
+++ b/index.html
@@ -351,30 +351,6 @@
             -
           </dd>
           <dt>
-            readonly attribute string format
-          </dt>
-          <dd>
-            -
-          </dd>
-          <dt>
-            readonly attribute string units
-          </dt>
-          <dd>
-            -
-          </dd>
-          <dt>
-            readonly attribute float near
-          </dt>
-          <dd>
-            -
-          </dd>
-          <dt>
-            readonly attribute float far
-          </dt>
-          <dd>
-            -
-          </dd>
-          <dt>
             readonly attribute Uint16Array data
           </dt>
           <dd>
@@ -515,6 +491,30 @@
           </dd>
           <dt>
             double verticalFieldOfView = null
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            double format = null
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            double units = null
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            double near = null
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            double far = null
           </dt>
           <dd>
             -
--- a/index.html
+++ b/index.html
@@ -544,56 +544,6 @@
         Examples
       </h2>
       <h3>
-        2D Canvas Context based post-processing
-      </h3>
-      <pre class='example highlight'>
-var canvasContext = document.createElement("canvas").getContext("2d");
-
-navigator.mediaDevices.getUserMedia({
-  video: true,
-  depth: true,
-}).then(function (stream) {
-  // wire the stream into a &lt;video&gt; element for playback
-  var video = document.querySelector('#video');
-  video.srcObject = stream;
-  video.play();
-  // wire the depth stream into another &lt;video&gt; element to convert kind
-  // NOTE: Only the R and G bytes are set to carry 16 bits of data
-  var depthVideo = document.querySelector('#depthVideo');
-  // construct a new MediaStream out of the existing depth track(stream)
-  var depthStream = new MediaStream(stream.getDepthTracks()[0]);
-  depthVideo.srcObject = depthStream;
-  depthVideo.play();
-
-  depthVideo.onloadedmetadata = function () {
-    function doSomething() {
-      canvasContext.drawImage(video);
-      var rgbImageData = canvasContext.getImageData(0, 0, w, h);
-      var pixels = rgbImageData.data;
-      canvasContext.drawImage(depthVideo);
-      var depthImageData = canvasContext.getImageData(0, 0, w, h);
-      var dexels = depthImageData.data;
-
-      // iterate through depth pixels to convert 2 bytes into 1 Uint16
-      for (var x = 0; x &lt; w ; ++x) {
-        for (var y = 0; y &lt; h; ++y) {
-          var i = (x + y * w) * 4;
-          // combine the R &amp; G pixels at (x, y) to get
-          // the 16 bit depth pixel value
-          var depth = dexels[i] | dexels[i + 1] &lt;&lt; 8;
-        }
-      }
-      // do things with pixels and dexels here
-      window.requestAnimationFrame(doSomething);
-    }
-    window.requestAnimationFrame(doSomething);
-  };
-}).catch(function (reason) {
-  // handle gUM error here
-});
-      
-</pre>
-      <h3>
         WebGL Fragment Shader based post-processing
       </h3>
       <pre class='example highlight'>
--- a/index.html
+++ b/index.html
@@ -316,45 +316,6 @@
       </section>
       <section>
         <h2>
-          <code>ImageData</code> interface
-        </h2>
-        <div class="note">
-          <p>
-            Depth cameras usually produce 16-bit depth values per pixel.
-            However, the canvas drawing surface used to draw and manipulate 2D
-            graphics on the web platform does not currently support 16bpp.
-          </p>
-          <p>
-            To address the issue, this specification defines a new data
-            representation for current <a>Canvas Pixel
-            <code>ArrayBuffer</code></a> of <code><a>ImageData</a></code>
-            interface to represent the 16bpp depth image produced by depth
-            cameras.
-          </p>
-        </div>
-        <p>
-          An <code><a>ImageData</a></code> object is said to <dfn>represent
-          depth data</dfn>, when the <a><code>CanvasImageSource</code></a> used
-          as the image source for the
-          <a><code>CanvasRenderingContext2D</code></a> is a <a>depth video
-          element</a>.
-        </p>
-        <p>
-          When representing a depth image, the <dfn>Canvas Pixel
-          <code>ArrayBuffer</code></dfn> is an <a><code>ArrayBuffer</code></a>
-          whose data is represented in left-to-right order, row by row top to
-          bottom, starting with the top left, with each pixel's lower 8 bit of
-          16 bit depth value, upper 8 bit of 16 bit depth value, 8 bit reserved
-          data, and another 8 bit reserved data being given in that order for
-          each pixel. Each component of each pixel represented in this array
-          must be in the range 0..255, representing the 8 bit value for that
-          component. The components must be assigned consecutive indices
-          starting with 0 for the top left pixel's lower 8 bit of 16 bit depth
-          value component.
-        </p>
-      </section>
-      <section>
-        <h2>
           <code>Settings</code> dictionary
         </h2>
         <p>
--- a/index.html
+++ b/index.html
@@ -316,6 +316,74 @@
       </section>
       <section>
         <h2>
+          <code>DepthData</code> interface
+        </h2>
+        <div class="note">
+          <p>
+            Depth cameras usually produce 16-bit depth values per pixel.
+            However, neither the canvas drawing surface used to draw and
+            manipulate 2D graphics on the web platform nor the
+            <code><a>ImageData</a></code> interface used to represent image
+            data support 16 bpp.
+          </p>
+          <p>
+            To address the issue, this specification defines a new
+            <code><a>DepthData</a></code> interface.
+          </p>
+        </div>
+        <dl class="idl" title="interface DepthData">
+          <dt>
+            readonly attribute unsigned long width
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute unsigned long height
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute string type
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute string format
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute string units
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute float near
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute float far
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute Uint16Array data
+          </dt>
+          <dd>
+            -
+          </dd>
+        </dl>
+      </section>
+      <section>
+        <h2>
           <code>Settings</code> dictionary
         </h2>
         <p>
--- a/index.html
+++ b/index.html
@@ -345,7 +345,7 @@
             -
           </dd>
           <dt>
-            readonly attribute string type
+            readonly attribute string measureType
           </dt>
           <dd>
             -
--- a/index.html
+++ b/index.html
@@ -421,6 +421,47 @@
       </section>
       <section>
         <h2>
+          <code>FrameGrabber</code> interface
+        </h2>
+        <dl class="idl" title=
+        "[Constructor(MediaStream stream)] interface FrameGrabber">
+          <dt>
+            readonly attribute MediaStream stream
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            void start()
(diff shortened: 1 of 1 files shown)


anssiko:
This fixes #76, #73, #72, and #66.

ds-hwang:
looks nice! lgtm

index.html
```diff
@@ -358,66 +316,149 @@
       </section>
       <section>
         <h2>
-          <code>CanvasImageSource</code> typedef
+          <code>DepthData</code> interface
```
huningxin: Would you consider to use DepthMap name? As depthmap (http://en.wikipedia.org/wiki/Depth_map) is a quite common name used in 3D camera space, e.g. Google Tango Project (https://developers.google.com/depthmap-metadata/) and Intel RealSense SDK are all using it in developer manual.

index.html
```diff
@@ -503,53 +544,33 @@
...
+  video: true
+}).then(function (mediaStream) {
+  // mediaStream carries both video track and depth track
+  var frameGrabber = new FrameGrabber(mediaStream);
+  if (frameGrabber) {
+    frameGrabber.start(processDepthData);
+  }
 });
+
+function processDepthData(frameData) {
```
huningxin: Would `processFrameData` be a better name?

index.html
```diff
@@ -358,66 +316,149 @@
...
-          data, and another 8 bit reserved data being given in that order for
-          each pixel. Each component of each pixel represented in this array
-          must be in the range 0..255, representing the 8 bit value for that
-          component. The components must be assigned consecutive indices
-          starting with 0 for the top left pixel's lower 8 bit of 16 bit depth
-          value component.
-        </p>
+        <dl class="idl" title="interface FrameData">
+          <dt>
+            readonly attribute DepthData? depthData
```
huningxin: Personally, I prefer to `DepthMap depthMap`

index.html
```diff
@@ -358,66 +316,149 @@
...
+            -
+          </dd>
+          <dt>
+            readonly attribute float far
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute Uint16Array data
```
huningxin: Probably, we need to define the algorithm about the meaning of the depth value. We may refer to https://developers.google.com/depthmap-metadata/encoding.

index.html
```diff
@@ -358,66 +316,149 @@
...
+            -
+          </dd>
+          <dt>
+            readonly attribute unsigned long height
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute string type
```
huningxin: Probably, we need a more accurate name, e.g. "measureType". And we need to define the valid values for this attribute,  e.g. "OpticalAxis" and "OpticRay".

index.html
```diff
@@ -358,66 +316,149 @@
...
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute string format
```
huningxin: We need to define the valid values for this attribute, say "RangeInverse" and "RangeLinear".

index.html
```diff
@@ -358,66 +316,149 @@
...
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute string units
```
huningxin: We need to define the valid values for this attribute, say "m" and "mm".

huningxin:
Thanks, @anssiko ! It looks nice. My comments provided.

index.html
robman: That's a fair point - linguistically I like the symmetry of ImageData/DepthData - but since there's an existing convention I'd be happy if we ended up with ImageData/DepthMap.

index.html
robman: +1 I'd buy that

anssiko:
@huningxin @robman @ds-hwang Thanks for the review and comments. I've addressed them in an update to this PR.

Some questions on the new proposed `DepthMap` attributes:
- Why wouldn't we provide the `data` in mm units directly? With Uint16Array  0..65535 we could represent dexels in the range of 0..65.535 metres which is probably enough. No need for `units` attribute.
- It seems a web developer would just prefer to indicate whether more precision is needed on the far end (think measuring distance to far away objects) or near (think face recognition), and then let the implementation take care of choosing whatever format under the hood that provides the best precision. The encoding pipeline (e.g. https://developers.google.com/depthmap-metadata/encoding) would be an implementation detail. The web developer would just get the `data` in mm units. This would allow us to drop `format`, `near`, and `far` since they are only needed by the implementation to implement the  encoding pipeline.
- What is the use case for `measureType`? The distance measured along the optical axis is longer, so I assume with that information a web developer can get more precise depth measurement. What other data the web developer would need in order to make use of `measureType` for real (is what we have defined in `Settings` enough)?

robman:
Hi @anssiko 

For your second point, I don't believe we should drop near/far/format. Near and far variables required to reconstruct depth if you receive a normalised depth value. Basically format defines which algorithm to use.

For your last point, the measureType defines if the length is measured along the optical axis (depth down the X3 axis in the pinhole camera model) or as a ray (along the effective hypotenuse). Again this tells us what algorithm to use.

anssiko:
@robman Thanks, what are the key use cases for which it would be preferred to use of the raw depth map data over the normalized data? And respectively, what are the key use case that require normalized data? I'm interested in figuring out how much of the details we can defer to the implementation, and how much we must expose to address the requirements. We should not expose metadata that has no supporting use case, and minimize the API surface instead, even if exposing metadata would be cheap.

Do the known implementations support both the measureTypes?

huningxin:
Hi @anssiko

Thanks for the updating.

> Why wouldn't we provide the data in mm units directly? With Uint16Array 0..65535 we could represent dexels in the range of 0..655.35 metres which is probably enough. No need for units attribute.

It sounds good to me. Web platform can stick to one "native" unit and data format. Implementations figure out how to fit into it. There are two flavors:
1. Uint16 in mm (BTW, the the range would be 0..65.535 m, correct?)
2. Float32 in m

What do you like?

> This would allow us to drop format, near, and far since they are only needed by the implementation to implement the encoding pipeline.

To encoding the depth value, e.g. implement https://developers.google.com/depthmap-metadata/encoding on web, web developers need to know near and far.

> What is the use case for measureType?

Agree with @robman , it tells web developer the depth measurement model, some algorithms depend on it.

`near`, `far`, `measureType` can be in MST's `Settings`.

anssiko:
@ds-hwang pointed out that Uint16 maps better to GPU when uploading this data to a WebGL texture. I'll let @ds-hwang expand.

@huningxin Yes, using Uint16 with mm units would represent the range of 0..65.535 m as you say.

If we'd stick with a single format (say range linear) and units (say mm) what use cases would we miss? At least that'd make the API simpler for the web developer. IOW, what are the key use cases that require the web developer to be able to convert the normalised depth values back?

ds-hwang:
Uint16 has enough accuracy. Let me explain.

First of all, we need to support RangeLinear mode and RangeInverse mode like https://developers.google.com/depthmap-metadata/encoding

User will calculate real distance using depth camera output.
Let say depth value is 10000 on [0, 65535], near is 1m, far is 5m.

RangeLinear formula is RealDistance = d(far - near) + near
so, 10000/65535 \* (5m - 1m) + 1m is real distance. 65535 step is accurate enough

RangeInverse formula is RealDistance = (far x near)/(far - d*(far - near))
user can calculate real distance.

As you see, depth value is just scale. it don't has unit. near and far has unit.

RangeLinear is used for dance game or something and RangeInverse is used for face recognition.

On the other hands, chromium keeps the cam video in texture. In the same sense, chromium will keep depth value in texture. 32float texture is supported on only extremely modern gpu. IMO 32float texture is overkill.

huningxin:
IMO, with Uint16 with mm units (or Float32 with m units),  we won't need to support encoding format, say RangeLinear and RangeInverse. The Uint16 depth value represents the real distance, e.g. 1 means 1 mm and 65535 means 65535 mm. It is straightforward to web developers, just use the value without any calculations. The only concern is that whether the range and accuracy is too limited, say will 65.535 m be too small or mm units is too large for some use cases with some new depth cameras in the future?

I propose to keep near and far as we may need to support web app to encode the depth value into other formats. Just like https://developers.google.com/depthmap-metadata/encoding, if web developer wants to encode the depth metadata into XMP properties, they need the near and far values. They would write JavaScript code to implement either RangeLinear or RangeInverse encoding mechanism.

ds-hwang:
I trust your judgement. However, I have to say there are drawback with value with unit.

Let's assume near is 0.1m and far is 5m. (0, 100) and (5000, 65535) are useless at that time. It's why tango encoding format uses scale value, instead of real value.

Some depth sensor in the future can have capability to detect >65m range. Some application want more precision <1mm.

If value has unit, it looks not flexible.

What is real output of RealSense and Kinect?

anssiko:
I updated the PR. Please review and comment. The use case for `measureType` was unclear so I dropped it for now. Also unclear if all implementations are able to support it.

I'd like to get resolution on the issue whether we should just hardcode `format` and `unit`, or allow flexibility. We should investigate the capabilities of the existing implementations and hardware to make that call.

If we keep `format`, I'd guess it'd make sense to allow a web developer to configure the `format`, right? Hypothetical API:

```
partial dictionary MediaStreamConstraints {
    // ...
    (DOMString or MediaTrackConstraints) depthFormat = "linear";
};
```

huningxin:
> What is real output of RealSense and Kinect?

RealSense SDK:

| Format | Description |
| --- | --- |
| PIXEL_FORMAT_DEPTH | The depth map data in **16-bit unsigned integer**. The values indicate the distance from an object to the camera's XY plane or the Cartesian depth.The value precision is in **millimeters**. |
| PIXEL_FORMAT_DEPTH_RAW | The depth map data in **16-bit unsigned integer**. The value precision is **device specific**. The application can get the device precision via the QueryDepthUnit function; |
| PIXEL_FORMAT_DEPTH_F32 | The depth map data in **32-bit floating point**. The value precision is in **millimeters**. |

Kinect SDK:
https://msdn.microsoft.com/en-us/library/microsoft.kinect.kinect.idepthframe.aspx
The data for this frame is stored as **16-bit unsigned integers**, where each value represents the distance in **millimeters**. The maximum depth distance is 8 meters, although reliability starts to degrade at around 4.5 meters. Developers can use the depth frame to build custom tracking algorithms in cases where the IBodyFrame isn’t enough.

Project Tango:
https://developers.google.com/project-tango/overview/depth-perception#point_clouds
The Project Tango APIs provide a function to get depth data in the form of a point cloud. This format gives (x, y, z) coordinates for as many points in the scene as are possible to calculate. Each dimension is a **floating point value** recording the position of each point in **meters** in the coordinate frame of the depth-sensing camera.

> Some depth sensor in the future can have capability to detect >65m range. Some application want more precision <1mm

This is why Float32 with meter units seems promising.

huningxin:
Point Cloud Library (PCL) is using **float** in RangeImage (depth map):
http://docs.pointclouds.org/trunk/classpcl_1_1_range_image.html

with **millimeters** units:
http://docs.pointclouds.org/trunk/classpcl_1_1_image_grabber_base.html#a32ae91b66b415213ec3b7c29d7e61e49

index.html
```diff
@@ -358,66 +317,211 @@
...
+            -
+          </dd>
+          <dt>
+            readonly attribute unsigned long height
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            readonly attribute Uint16Array data
```
huningxin: I propose to define `data` as `Float32Array` and specify the units is millimeters.

anssiko:
@huningxin Thanks for sharing information on implementations. What is your guesstimate re the performance implications (also memory) of using `Float32Array` over `Uint16Array`? I think we should expect frame rates of >=30 Hz. I guess some benchmark data would help to make an informed decision. @ds-hwang had some concerns, but we don't have benchmark data at hand now.

If there are performance concerns, one approach worth considering might be to go with the lowest common denominator (e.g. `Uint16Array`, mm units) first, while allow future extensions. For example, a new `MediaTrackConstrains` could be used to indicate higher precision is preferred, and the type of `data` could be updated to `(Uint16Array or Float32Array)` while keeping the API backwards compatible. Not optimal considering interoperability, but I think that might be a reasonable tradeoff to make.

Re units, mm sounds like the best choice regardless of the type.

huningxin:
@anssiko , thanks for the comments. I agree with you that the `Uint16Array` seems to be closer to hardware.

Because:
In RealSense SDK, the depth map data is in 16-bit unsigned integer of `PIXEL_FORMAT_DEPTH_RAW` format. The units is device defined.

Kinect SDK uses 16-bit unsigned integers with mm

According to Tango Depth Camera implementation in Chromium (https://code.google.com/p/chromium/codesearch#chromium/src/media/base/android/java/src/org/chromium/media/VideoCaptureTango.java&q=tango&sq=package:chromium&l=141),

```
    // Depth is composed of 16b samples in which only 12b are
    // used.
```

It is also `uint16`.

So `Uint16Array` with `mm` units looks good to me.

For the units, I suggest we add `long units` into `MediaTrackConstrains`.

index.html
```diff
@@ -453,26 +557,63 @@
           <dd>
             -
           </dd>
+          <dt>
+            DOMString format = null
```
huningxin: I think we don't need `format` anymore.

index.html
```diff
@@ -453,26 +557,63 @@
...
+            -
+          </dd>
+          <dt>
+            DOMString unit = null
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            double near = null
```
huningxin, line 580: As depth value is in `Uint16Array`, `near` should be `short`, is it?

index.html
```diff
@@ -453,26 +557,63 @@
...
+          </dt>
+          <dd>
+            -
+          </dd>
+          <dt>
+            double far = null
```
huningxin, line 586: As depth value is in `Uint16Array`, `far` should be `short`, is it?

index.html
anssiko: Dropped.

index.html
anssiko, line 580: The Web Platform APIs use `double` even if `short` would be more appropriate. Probably due to the fact that `double` maps so nicely to ECMAScript `Number`:

http://heycam.github.io/webidl/#es-double

I suggest we stick with `double`.

index.html
anssiko, line 586: Ditto.

anssiko:
@huningxin Thanks again for your suggestions. I updated the spec to address your comments. I also did further refactoring. I'd like to review this after your final review.

index.html
huningxin, line 580: lgtm. Thanks for the explanation!

index.html
huningxin, line 586: lgtm.

index.html
```diff
@@ -276,9 +240,54 @@
...
+          <dt>
+            DepthMapUnit unit = "mm"
+          </dt>
+          <dd>
+            -
+          </dd>
+        </dl>
+        <dl id="enum-basic" class="idl" title="enum DepthMapUnit">
+          <dt>
+            mm
```
huningxin, line 262: Will people confuse 'mm' between millimeters and micrometers?

huningxin:
hi @anssiko , thanks for your efforts and explanation. It makes the spec pretty good.

LGTM with one open.

index.html
anssiko, line 262: Millimeter (the American spelling of millimetre) has an SI unit symbol 'mm' so I think that's the best we can do. If we'd use the full name (millimeter) people would probably typo it since the American and International spellings differ subtly.

anssiko:
@huningxin @robman @ds-hwang I'll merge this PR now and craft a mail to group to get wider feedback. Thanks for your contributions and review!

index.html
huningxin, line 262: It sounds good to me. Thanks for the explanation!

Issue #75, opened 2015-02-27
Update implementation considerations re track synchronization and calibration 


anssiko:
Currently we have the following non-normative section in the spec: http://w3c.github.io/mediacapture-depth/#implementation-considerations

> A MediaStreamTrack object representing a video track and a MediaStreamTrack object representing a depth track can be combined into one MediaStream object. The rendering of the two tracks are intended to be synchronized. The resolution of the two tracks are intended to be same. And the coordination of the two tracks are intended to be calibrated. These are not hard requirements, since it might not be possible to synchronize tracks from sources.

With the currently defined API we can create a depth-only MediaStream `s2` out of the combined "RGB and depth" MediaStream `s` and assign `s2` to `<video>` element `v` as follows:

```
var s2 = new MediaStream(s.getDepthTracks()[0]);
v.src = URL.createObjectURL(s2);
```

Any implementability issues in keeping the RGB stream synchronized and calibrated with the depth stream if they are combined into a single MediaStream versus split into two MediaStreams as long as they originate from the same MediaStream initially?

Referring back to https://www.w3.org/wiki/Media_Capture_Depth_Stream_Extension#Requirements it seems we attempt to address LMD4 and LMD5 but only for a combined MediaStream.

Can we rephrase implementation considerations to apply to cases in which the combined stream is split into two MediaStreams? IOW as long as the tracks originate from the same MediaStream best effort would be taken to keep them synchronized and calibrated. I think this would be the expectation of web developers, but may be hard to implement as it requires coordination between MediaStreams.

ds-hwang:
First of all, we must need to support split into two MediaStreams, because some cases only need depth stream and others only need video stream. For example, gesture control needs only depth stream, and someone uses RealSense as just cam.
Furthermore, we want it possible to draw depth stream on video element, which means some cases want to make video element draw depth stream and others  want to make video element draw video stream.

However, combined single VideoElement is not necessary. It provides just additional convenience. I'm not fan of combined single VideoElement because
1. visual result of combined single VideoElement is just same to VideoElement based on video stream
2. It requires additional implementation changes. Blink video code must deal with combined video, just video, and depth video.

I think combined single VideoElement doesn't require huge architecture change. It just add additional complexity to code. It might be better to postpone to next stage.

Synchronization might bring up combined single VideoElement. It's straightforward to synchronize video stream VideoElement and depth stream VideoElement because the source is same, which means the video and depth frame is decoded together.
[...]

PR #86, opened 2015-11-20
Define extensions for media provider object and the video element, conversion to grayscale

anssiko:
This is the first stab based on discussion in https://github.com/w3c/mediacapture-depth/issues/75

PTAL @huningxin @bbernhar. We can further incubate this before landing.


huningxin:
LGTM. Thanks for this PR!

bbernhar:
LGTM.

PR #93, opened 2016-01-21
Setup Echidna for continuous deployment from TR branch

anssiko:
This is to close an [action](https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0031.html) to pilot the use of automated publication with Echidna with mediacapture-depth. The setup is as such that whenever there's a push to the TR branch, a new WD is published.

@dontcallmedom Please review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,19 @@
-language: python
-install:
- - git clone https://github.com/w3c/respec.git
- - git clone https://github.com/dontcallmedom/webidl-checker.git
- - git clone https://github.com/dontcallmedom/widlproc
- - cd widlproc && make obj/widlproc && cd ..
- - pip install html5lib lxml
+language: node_js
+
+sudo: false
+
+branches:
+  only:
+    - TR
+
+env:
+  global:
+    - URL="https://rawgit.com/w3c/mediacapture-depth/TR/ECHIDNA"
+    - DECISION="https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0031.html"
+    - secure: "NOnRV3zs4zT3OFN5lSnuzc//VxvRKcUmGH/4YPQM2Xqm/LorP4D35GtREJL3o/n273CxyCV6IRtx/LPOSQQymZVSwfMHfIB+QA+llaGw6eItVGq0thxXGncXv5z8L6XV6cT7B0Y5UEK5GaHUdzyCHT5L8YvnadeXlVsASBAMinI="
+ 
 script:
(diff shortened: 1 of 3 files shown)


dontcallmedom:
LGTM, although this removes the bits of the travis set up used to do actual continuous integration (although these bits aren't using the latest approach for that either, so probably not a big loss at this point). Also, for obvious reason, I can't verify that the publication token is correct :)

anssiko:
@dontcallmedom Thanks for the review. I think the best way to test whether this works as expected is to give it a spin and try to publish a new WD using the new setup. We can refine the CI steps in subsequent updates, add webidl-check and friends.

The .travis.yml used is based on https://github.com/w3c/echidna/wiki/Using-Travis-with-Echidna. Is there a more authoritative guide for using Travis with Echidna?

dontcallmedom:
https://github.com/w3c/echidna/wiki/Using-Travis-with-Echidna is the authoritative source (at least, that's the one I use myself :)

anssiko:
All seems good, except I forgot to add w3cids to the spec source for all the editors (@dontcallmedom let's continue debug this off the GH). Here's the Echidna error log https://lists.w3.org/Archives/Public/public-tr-notifications/2016Jan/0025.html for the curious.

.travis.yml
```diff
@@ -1,11 +1,19 @@
...
+
+sudo: false
+
+branches:
+  only:
+    - TR
+
+env:
+  global:
+    - URL="https://rawgit.com/w3c/mediacapture-depth/TR/ECHIDNA"
```
dontcallmedom, line 11: Not sure if that explains the bug we've encountered, but can you try with http://htmlpreview.github.io/?https://github.com/w3c/mediacapture-depth/TR/ECHIDNA instead? (the publication token will need to be updated in this case I think)

anssiko, line 11: The rawgit.com URL works (well, at least last time we published) for the Presentation API with Echidna, so I believe this cannot be the reason for the failure?

https://github.com/w3c/presentation-api/blob/gh-pages/.travis.yml#L11

dontcallmedom, line 11: you're right, forgot what I said, I got confused :)

anssiko:
Current status: we hit a bug in Echidna likely related to an upgrade of the backend, see https://github.com/w3c/echidna/issues/251

dontcallmedom:
The bug was fixed, and the publication went through this time: https://lists.w3.org/Archives/Public/public-tr-notifications/2016Jan/0030.html; yeah :)

anssiko:
Excellent, thanks for debugging this with me!

PR #95, opened 2016-02-22
Fix #94: Clarify constraint selection order

### LINKED ISSUE #94: Clarify constraint selection order from getUserMedia
If the UA requests with the constraints of `{video: true, audio: true, depth: true}`, it's not clear if the selected corresponding video device was decided by a paired default audio device and a separate depth device OR a paired depth device and a separate non-default audio device. 

Per 6.6.1, we should consider enforcing the order of selection with a depth constraint where it takes precedence in deciding a corresponding video device. This would prevent selecting a combined stream with separate group identifiers in implementations.

anssiko:
Fix #94.

### DIFF
--- a/index.html
+++ b/index.html
@@ -370,6 +370,12 @@
           null, the <a>MediaStream</a> MUST NOT contain a <a>depth stream
           track</a>.
         </p>
+        <div class="note">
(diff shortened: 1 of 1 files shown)


anssiko:
Hearing no concerns -- merging.

PR #108, opened 2016-03-15
Fix #99: Clarify handling of depth as object in MediaStreamConstraints

### LINKED ISSUE #99: Clarify handling of `depth` when set to an object
6.1 states 

> If the depth dictionary member has the value true, the MediaStream returned by the getUserMedia() 
> method MUST contain a depth stream track. If the depth dictionary member is set to false, is not 
> provided, or is set to null, the MediaStream MUST NOT contain a depth stream track. "

It doesn't say explicitly what to do when it is set to an object; the media capture and spec states 

> If a Constraints structure is provided, it further specifies the nature and settings of the [...] track

(I'm not sure that handling of the unit system as a constraint is a good use of constraints; but that'll come as a separate issue)

anssiko:
Fix #99

### DIFF
--- a/index.html
+++ b/index.html
@@ -354,7 +354,10 @@
           method MUST contain a <a>depth stream track</a>. If the <a>depth</a>
           dictionary member is set to false, is not provided, or is set to
           null, the <a>MediaStream</a> MUST NOT contain a <a>depth stream
-          track</a>.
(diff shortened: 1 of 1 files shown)


huningxin:
LGTM. Thanks for the PR!

PR #111, opened 2016-03-16
Fix #100: Depth precision in constraints and settings

### LINKED ISSUE #100: Setting unit in constraints
The way constraints for mediastreamtrack have been defined is to allow browser-mediated negotiations of allowable ranges in which an application can operate on the said track.

Using this to indicate the unit of the reported values seems a poor application of that concept, since there is no particular reason why one would negotiate between meters and milimeters (or at least, I can't see what reason); in general, I'm not sure there is any value in saving the developers from a division or multiplication by a fixed number (1000) here.

anssiko:
This is an attempt to Fix #100. See the issue https://github.com/w3c/mediacapture-depth/issues/100#issuecomment-197380058 for remaining opens that need to be addressed before we can land this and provide your feedback. Thanks.

### DIFF
--- a/index.html
+++ b/index.html
@@ -320,7 +320,7 @@
           view in degrees.
         </p>
         <p>
-          A <a>depth map</a> has an associated <dfn>unit</dfn> which is a
+          A <a>depth map</a> has an associated <dfn>precision</dfn> which is a
           string. It represents the <a>depth precision</a>.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -320,18 +320,17 @@
...
-          string. It represents the <a>active depth map unit</a>.
+          A <a>depth map</a> has an associated <dfn>precision</dfn> which is a
+          string. It represents the <a>depth precision</a> (one of the
+          <a>possible depth precisions</a>).
         </p>
         <p>
           A <a>depth map</a> has an associated <dfn>near value</dfn> which is a
-          double. It represents the minimum range in <a>active depth map
-          unit</a>s.
+          double. It represents the minimum range in millimeters.
```
jan-ivar, line 324: Wondering, would it make more sense to go with the base unit (meters) since milli is literally .001 and this number is double precision, or is millimeter a defacto enough standard here already that it would be weird?

anssiko, line 324: @huningxin WDYT? I'd prefer a unit for near and far values that makes the associated math simpler. The de facto unit for focal length is millimeters -- and that's what we use in this spec too.

huningxin, line 324: I'd prefer the `depthNear` and `depthFar` follow `depthMapUnit`. And as today's hardware commonly produce 16bit depth value in millimeter, so I prefer the default unit is `mm`.

PR #118, opened 2016-04-05
Fix #114: Add depthNear and depthFar constrainable properties

### LINKED ISSUE #114: Range constraints
Creating a standalone issue for this discussion started in #100:

> @jan-ivar:
> I naively imagine I'd probably want a different setting for pointing my kinect out my window vs. pointing it at my couch?
> 
> @martinthomson:
> @jan-ivar, that would be a separate constraint, probably two.  minimumRange: 1, maximumRange: 100.  If that's a real concern, then a separate issue.  For this one, I think that @anssiko has the right approach: find out how do the devices actually measure and work from there.
> 
> @huningxin:
> @jan-ivar, Kinect SDK allows to configure depth camera range (`Default` or `Near`) [[1]](https://msdn.microsoft.com/en-us/library/microsoft.kinect.depthimagestream.range.aspx). libealsense [[2]](https://github.com/IntelRealSense/librealsense/blob/master/include/librealsense/rsutil.h#L107) allows to configure several ranges to RealSense front facing depth camera as well (`Common`, `ShortRange`, `LongRange` etc.,).  I am not aware Tango SDK allows to configure working range of its depth camera.

anssiko:
As a fix to #114 this PR adds `depthNear` and `depthFar` constrainable properties. Note we switch to use meter as the unit (from millimeters) for the near value and far value as suggested by @martinthomson, and also the recovered depth measurement `d` is now reported in meters for consistency.

I have one question for the group:
- Do we want to special case `depthFar < depthNear`, or leave it to the implementation to figure out what to do in such a case?

Seeking review from @huningxin @martinthomson @jan-ivar @alvestrand et al.

### DIFF
--- a/index.html
+++ b/index.html
@@ -492,8 +492,6 @@
           given a depth map value <var>d</var>, is as follows:
         </p>
         <ol>
-          <li>Let <var>bit depth</var> be the bit depth of the depth map.
-          </li>
           <li>Let <var>near</var> be the the <a>near value</a>.
           </li>
           <li>Let <var>far</var> be the the <a>far value</a>.
--- a/index.html
+++ b/index.html
@@ -224,8 +224,6 @@
         The <code><a href=
         "https://w3c.github.io/mediacapture-main/#dom-mediadevices-getusermedia">
         <dfn>getUserMedia()</dfn></a></code>, <code><a href=
-        "https://w3c.github.io/mediacapture-main/#dfn-applyconstraints"><dfn>applyConstraints()</dfn></a></code>,
-        <code><a href=
         "https://w3c.github.io/mediacapture-main/#dfn-getsettings"><dfn>getSettings()</dfn></a></code>
         methods and the <code><a href=
         "https://w3c.github.io/mediacapture-main/#idl-def-NavigatorUserMediaSuccessCallback">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -583,6 +581,85 @@
...
+              <td>
+                The <a>far value</a>, in meters.
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <p>
+          The <a>applyConstraints()</a> method MUST reject the promise with
+          <code>OverconstrainedError</code>, when invoked with
+          <code>depthNear</code> or <code>depthFar</code> property.
```
martinthomson: I don't think that this is necessary.  Changing near and far for the purposes of changing where the 8-bits are applied seems reasonable.  You can maybe note that changing the operating mode of the camera might not be possible, which would result in an `OverConstrainedError` when the near and far constraints are outside the operating range.  However, as long as near and far are possible, I see no reason to prohibit this.  (Feel free to push this to a followup.)

martinthomson:
Oh, and depthFar < depthNear => OverConstrainedError.

index.html
anssiko: Right. That behavior is already defined in mediacapture-main [`applyConstrains()`](https://w3c.github.io/mediacapture-main/#widl-ConstrainablePattern-applyConstraints-Promise-void--Constraints-constraints) so I removed it, see 13132ae. We do not need to patch the `applyConstrains()` algorithm in this spec.

anssiko: For the `depthFar < depthNear` case, I guess we can lean on the `overconstrained` concept defined in mediacapture-main, see 2835955.

anssiko:
@martinthomson Thanks for the feedback! Please take a look at 13132ae and 2835955 commits in this branch that should address your concerns.

All - please let us know if you have concerns with the suggested changes in this PR. I'd like to merge this soonish.

Issue #113, opened 2016-03-23
Support a consistent measurement unit


martinthomson:
We support i420 and RGB for video without burdening applications with the distinction between the two.  In the browser itself, those two might be carried differently, but when they hit a screen or API, they are converted to a common form.

It would appear that depth captures aren't consistent in the units they use.

I can understand a representation that was expressed relative to the focal length or some other physical constraint if the intent is to preserve the maximum dynamic range over the 8 or 16 bits available to the encoding, but the spec already does that in terms of the near and far value.

The near value and far value in particular could be expressed in metres.

anssiko:
#111 is consistent in depth value representation, and standardizes on millimeters. I'm hearing this is the unit that the native APIs use to represent depth.

anssiko:
We eventually settled on meters (using this U.S. English spelling in the spec assuming the [W3C Manual of Style](https://www.w3.org/2001/06/manual/#Translations) is to be obeyed still).

Specifically, fixed this issue with 13c5bda in #118, as discussed in #114.

PR #109, opened 2016-03-15
Fix #103: Rename depth-specific MediaTrackSettings and enums

anssiko:
Fix #103:

Rename depth-specific MediaTrackSettings:

attribute `format` -> `depthConversion`
attribute `near` -> `depthNear`
attribute `far` -> `depthFar`

Rename the enums to give them unique names:

enum `RangeFormat` -> `DepthConversionEnum`
enum `DepthMapUnit` -> `DepthMapUnitEnum`

Issue #103, opened 2016-03-08
`format` is too generic for `MediaTrackSettigs`


dontcallmedom:
`MediaTrackSettings`, for better or for worse, will have to host all the possible settings for audio, video (and based on this spec, depth) tracks.

Using a term as generic as "format" for something so specific to depth tracks seems wrong.

jan-ivar:
`grayscaleConversion`?

PR #119, opened 2016-04-07
Fix #102: Define non-constrainable property

### LINKED ISSUE #102: Settings vs constrainable properties
The way Media Capture and Streams currently operates, the set of constrainable properties is the same as the set of settings obtained via getSettings().

This spec uses settings to report on non-constrainable properties.

That might be OK, but since that's clearly something that hasn't been done before, it would be good to check that this is consistent with the rest of the contrainable model.

anssiko:
- Define non-constrainable property
- Apply the newly minted definition to:
  - focalLength
  - horizontalFieldOfView
  - verticalFieldOfView

### DIFF
--- a/index.html
+++ b/index.html
@@ -692,12 +692,6 @@
           lens may be fixed, or the underlying platform may not expose the
           focal length information.)
         </p>
-        <p>
-          If the <a>near value</a>, <a>far value</a>, <a>focal length</a>,
-          <a>horizontal field of view</a>, or <a>vertical field of view</a> is
-          unknown due to any reason, the corresponding constrainable property's
-          value MUST be set to the value 0.
-        </p>
         <pre class="idl">
           partial dictionary MediaTrackConstraintSet {
             ConstrainDouble depthNear;
--- a/index.html
+++ b/index.html
@@ -224,8 +224,6 @@
         The <code><a href=
         "https://w3c.github.io/mediacapture-main/#dom-mediadevices-getusermedia">
         <dfn>getUserMedia()</dfn></a></code>, <code><a href=
-        "https://w3c.github.io/mediacapture-main/#dfn-applyconstraints"><dfn>applyConstraints()</dfn></a></code>,
-        <code><a href=
         "https://w3c.github.io/mediacapture-main/#dfn-getsettings"><dfn>getSettings()</dfn></a></code>
         methods and the <code><a href=
         "https://w3c.github.io/mediacapture-main/#idl-def-NavigatorUserMediaSuccessCallback">
@@ -240,9 +238,7 @@
         and <code><a href=
         "https://w3c.github.io/mediacapture-main/archives/20160406/getusermedia.html#event-mediastreamtrack-overconstrained">
         <dfn>overconstrained</dfn></a></code> as applied to
-        <a>MediaStreamTrack</a> as well as the concept <a href=
-        "https://w3c.github.io/mediacapture-main/#dfn-fitness-distance"><dfn>fitness
-        distance</dfn></a> are defined in [[!GETUSERMEDIA]].
+        <a>MediaStreamTrack</a> are defined in [[!GETUSERMEDIA]].
       </p>
       <p>
         The terms <a href=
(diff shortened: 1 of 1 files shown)


anssiko:
Fixes #102.

index.html
```diff
@@ -560,15 +564,18 @@
         <div dfn-for="MediaTrackSettings">
           <p>
             The <dfn><code>focalLength</code></dfn> dictionary member
-            represents the <a>depth map</a>'s <a>focal length</a>.
+            represents the <a>depth map</a>'s <a>focal length</a>. It is a
+            <a>non-constrainable property</a>.
```
jan-ivar: `non-constrainable` seems like a misnomer. Constraints are properties of the _application_, as in "here are my constraints", a request. So it is perfectly reasonable to set constraints on inherent unchangeable properties. Remember `getUserMedia` may pick from several cameras/mics.

index.html
```diff
@@ -661,6 +668,33 @@
...
         <h2>
+          Non-constrainable properties
+        </h2>
+        <p>
+          A <dfn lt=
+          "non-constrainable property|non-constrainable properties">non-constrainable
+          property</dfn> is ignored in the context of the
+          <a>applyConstraints()</a> algorithm. Its value can be obtained via
+          the <a>getSettings()</a> accessor. A <a>non-constrainable
+          property</a>'s <a>fitness distance</a> is always 0.
```
jan-ivar: This new thing is not in the constrainable algorithm today, and seems unnecessary. I see nothing harmful in people putting constraints on fixed values. Worst case `applyConstraints` just fails.

index.html
anssiko: I copied the name from #102 :)

Would rename of the concept into "unchangeable property" make things better? I'm happy to refactor the PR further given concrete suggestions how to convey the read-only-ness of these in the context of the depth stream tracks. Currently, those constraints do not exist for the regular MSTs, but could be if upstreamed to a future version of the mediacapture-main.

jan-ivar: Having things that only show up in getSettings violates the Constrainable pattern, so unless we have a good reason, I think we should mirror facingMode. Is there a downside?

jan-ivar: But yeah you're right, I'm objecting to the concept, not the name.

anssiko: I'd be happy to mirror it, if educated how to do so the right way. There were no hooks to related concepts that could be referenced, also the facingMode prose seemed not reusable, consider:

> This string (or each string, when a list) should be one of the members of VideoFacingModeEnum. The members describe the directions that the camera can face, as seen from the user's perspective. Note that getConstraints may not return exactly the same string for strings not in this enum. This preserves the possibility of using a future version of WebIDL enum for this property.

... and:

> A camera can report multiple facing modes. For example, in a high-end telepresence solution with several cameras facing the user, a camera to the left of the user can report both "left" and "user".

Concrete suggestions welcome!

index.html
anssiko: Ok, so will drop the fitness distance special casing.

index.html
jan-ivar: Or just mirror any other constraint, like `frameRate` (a numerical one is probably closer to what you want). There's nothing special with facingMode in this regard. I.e. add to:
- MediaTrackSupportedConstraints
- MediaTrackCapabilities
- MediaTrackConstraintSet
- MediaTrackSettings

(These four places used to be a single dictionary, but people who are not me decided to split them up).

anssiko:
@jan-ivar Thanks for the good suggestions! I reverted 1c8f97d and did another attempt to define the constrainable properties per your suggestion. Please review 66fab9e.

anssiko:
@jan-ivar Thanks for the feedback!
- The unit discussion happened in #113. Meter seem like a more natural unit for the near and far values, while millimeter is the de facto unit for the focal length. I think this is a reasonable tradeoff to make. In the common use cases you'd use near and far but not focal length. In advanced use cases, the web developers know how to convert between the units.
- The names were initially `near` and `far`. We prefixed them with `depth` based on feedback, see #109.

Changes:
- Removed the use of a special value (`0`) for constrainable properties in 3bb0bbe
- Used `(double or DoubleRange)` for `MediaTrackCapabilities` in 31ef274

Unless other comments arise, I'd like to land this PR.

jan-ivar:
Sorry I don't see any feedback in https://github.com/w3c/mediacapture-depth/pull/109. All I see is @dontcallmedom saying in https://github.com/w3c/mediacapture-depth/issues/103 that `format` is too generic, with which I agree. I don't think that criticism extends to `near` and `far` though or he would have mentioned it I think. I don't think he meant to blanket-prefix constraints with their domain, as we've never done that before. I can't think of another domain for `near` and `far` either (the only other domain I can think of would be time, but something like `timeNear` and `timeFar` would be a poor metaphor of distance applied time compared to better words like 'soon' and `late`), so I think we would be fine with `near` and `far` FWIW. Otherwise LGTM. @dontcallmedom thoughts?

anssiko:
@jan-ivar I dropped the depth prefix per your suggestion in b0b81df. Thanks for the review, landing now!

PR #120, opened 2016-04-13
Fix #104: Define associated permission for a depth camera source

### LINKED ISSUE #104: Integration with Permission API
The Media Capture and Streams spec has a pending pull request to integrate with the permission API; the latest draft of the Permission API describes [permissions for "cameras, microphones"](https://w3c.github.io/permissions/#media-devices).

I assume the permission for access to a depth map can be encompassed as part of the "camera" permissions; but that's probably worth clarifying both in this spec and possibly in the permissions API as well.

anssiko:
The Permissions API [says](https://w3c.github.io/permissions/#media-devices):

> The "camera", "microphone" , and "speaker" permissions are associated with permission to use media devices as specified in [GETUSERMEDIA] and [audio-output].

That suggests the "camera" permissions is generic, can be reused for depth. This PR defines that as such.

Fixes #104.

### DIFF
--- a/index.html
+++ b/index.html
@@ -262,6 +262,13 @@
         "https://html.spec.whatwg.org/#videotrack"><dfn>VideoTrack</dfn></a></code>
         interfaces are defined in [[!HTML]].
       </p>
+      <p>
+        The term <a href=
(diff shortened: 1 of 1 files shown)


anssiko:
Merging, can iterate in subsequent commits based on feedback if needed.

PR #122, opened 2016-04-21
Do not call echidna on pull requests

vivienlacourba:
It is not needed to call echidna when PR are made but only when they get merged.

FYI this comes from: https://github.com/w3c/echidna/wiki/Setting-up-Echidna-as-a-GitHub-hook
(of course this change will have have to be integrated on the TR branch)


anssiko:
Thanks!

PR #128, opened 2016-06-02
Make browser load the spec via file url.

ds-hwang:
This PR makes it possible to load this spec via file url. file:///<workspace>/mediacapture-depth/index.html


ds-hwang:
many places already use https://www.w3.org/XXX

anssiko:
Thanks! Now that W3C has migrated to https-only this seems fine. Was there for historical reasons.

Issue #110, opened 2016-03-16
Accessing Camera Calibrations


bbernhar:
I see the importance that we should consider allowing the UA to access calibration data. It would enable many useful computer vision scenarios and bring the need to enable at-least three calibrations:
- Spatial coordinate system: This gives us ability to project and unproject 2D image space to 3D camera space and map between locations of depth and the corresponding color image. Ideally, this wouldn't be done on a per-pixel basis and a coordinate system could represent the UA's world position too.
- Lens distortion: The UA can remove distortion from the image as not all cameras use high-distortion lenses.
- Camera re-sectioning: This can unproject 2D point to 3D space as ray (3D surface reconstruction from image or "skeletal tracking").
  
  The `v1` spec does not mention how or which calibration is used unlike the more verbose SDK-specific APIs [1] [2] and it seems necessary to expose these intrinsic operations and properties that can be performed on streams that belong to the same group of 3D devices.

@anssiko @huningxin WDYT?

[1] https://msdn.microsoft.com/en-us/library/dn785530.aspx
[2] https://software.intel.com/sites/landingpage/realsense/camera-sdk/v1.1/documentation/html/doc_essential_coordinates_mapping.html

anssiko:
@bbernhar Thanks for raising this issue. I believe this hinges on the question how to do the split between v1 and v2 features.

Could you craft a proposal (in plain text, as a comment to this issue) with the intrinsic operations and properties that form the intersection of the capabilities of the most popular devices that are widely shipping? I assume these would be exposed via `MediaTrackSettings`?

Also, explaining the use cases might be helpful too. Your three scenarios already shed some light, but it may not be clear to everyone what the concrete usage scenarios are we'd miss if we'd defer these to v2.

bbernhar:
Thanks @anssiko. This seems beyond the `v1` horizon but here are the basic problems I can see that need further discussion. There exist equivalent operations that ship for Kinect and RealSense, at-least (links below).

### Spatial Coordination

Uses calibrations from streams to reason with the related coordinate systems.

UC: Skeletal or 3D tracking of objects [1].

Detect where some object identified in depth space is in color space:

```
var objectPointsInDepth = ... // array of 2D points in depth space

//  Acquire & update per frame.
var depthCoordSys = depthStream.coordSys()
var colorCoordSys = colorStream.coordSys()

var objectsPointsInColor = depthCoordSys.map(objectPointsInDepth, colorCoordSys)
```

[...]

PR #130, opened 2016-06-02
Fix #125: define the data type of a depth map is 16-bit unsigned integer.

### LINKED ISSUE #125: How to handle depth conversion when range has a near value of zero.
The RSSDK `PROPERTY_DEPTH_SENSOR_RANGE` property [1] can return a value of zero for `MinDepth`. Using our _convert using range inverse_  (per 6.6), the depth map will only quanitize to `dn=1` when near=0 and will inaccurately compute the interpolated depth values when converted into the grayscale image.

This is because the depth range (i.e. `MinDepth` and `MaxDepth` properties) can be reported as "effective range" or "camera range". If the formula uses camera range (what the RSSDK uses), the conversion output is not usable depth.

Effective range (i.e. MinDepth >0) would require computation on the depth frame, since hard coding "ideal" values isn't a long-term solution. However, this would be rather expensive given one would need to re-compute the range every frame.

I don't have a good solution there; other than to avoid the conversion altogether (e.g. move away from 8-bit depth) or just hard code ideal "effective range" values so we could avoid always processing each pixel per frame. Alternatively, we could define the behavior to only compute the range once upfront and leave the computation to script. 

[1] https://software.intel.com/sites/landingpage/realsense/camera-sdk/v1.1/documentation/html/index.html?property_device_pxccapture.html

ds-hwang:
In #125, I proposed 
1. explicitly define 16-bit integer as depth type
2. explicitly declare this spec depends on WebGL2

Our friends agreed, so let me propose this PR.

Fix #125: define the data type of a depth map is 16-bit unsigned integer.
Fix #125: explicitly declare this spec depends on WebGL2.

### DIFF
--- a/index.html
+++ b/index.html
@@ -226,23 +226,9 @@
         or <a>depth+video stream</a>.
       </p>
       <p>
-        Depth cameras usually produce 16-bit depth values per pixel. However,
-        neither the canvas drawing surface used to draw and manipulate 2D
-        graphics on the web platform nor the <code><a>ImageData</a></code>
-        interface used to represent image data support 16 bits per pixel. To
-        address the issue, this specification defines a conversion into a 8-bit
-        grayscale representation of a <a>depth map</a> for consumption by APIs
-        that are limited to 8 bits per pixel.
-      </p>
-      <p>
-        The Media Capture Stream with Worker specification
-        [[MEDIACAPTURE-WORKER]] that complements this specification enables
-        processing of 16-bit depth values per pixel directly in a worker
-        environment and makes the <code>&lt;video&gt;</code> and
-        <code>&lt;canvas&gt;</code> indirection and depth-to-grayscale
-        conversion redundant. This alternative pipeline that supports greater
-        bit depth and does not incur the performance penalty of the indirection
-        and conversion enables more advanced use cases.
+        Depth cameras usually produce 16-bit depth values per pixel, so this
+        specification defines a 16-bit grayscale representation of a <a>depth
+        map</a>.
       </p>
     </section>
     <section>
@@ -414,170 +400,9 @@
           view</dfn> which is a double. It represents the vertical angle of
           view in degrees.
         </p>
-      </section>
-    </section>
-    <section>
-      <h2>
-        Extensions
-      </h2>
-      <section>
-        <h2>
-          <code><a>MediaStreamConstraints</a></code> dictionary
-        </h2>
-        <pre class="idl">
-          partial dictionary MediaStreamConstraints {
-              (boolean or MediaTrackConstraints) depth = false;
-          };
-        
-</pre>
-        <p dfn-for="MediaStreamConstraints" link-for="MediaStreamConstraints">
-          If the <dfn><code>depth</code></dfn> dictionary member has the value
-          true, the <a>MediaStream</a> returned by the <a>getUserMedia()</a>
-          method MUST contain a <a>depth stream track</a>. If the <a>depth</a>
-          dictionary member is set to false, is not provided, or is set to
-          null, the <a>MediaStream</a> MUST NOT contain a <a>depth stream
-          track</a>. If the <a>depth</a> dictionary member is set to a valid
-          <a>Constraints</a> dictionary, the <a>MediaStream</a> returned by the
-          <a>getUserMedia()</a> method MUST contain a <a>depth stream track</a>
-          that fulfills the specified mandatory constraints.
-        </p>
         <p>
-          The <a>permission</a> associated with a depth camera <a>source</a> is
-          "<a>camera</a>",
-        </p>
-        <div class="note">
-          If the user agent requests a combined <a>depth+video stream</a>, the
-          devices in the constraint should be satisfied as belonging to the
-          same group or physical device. The decision to select and satisfy
-          which device pair is left up to the implementation.
-        </div>
-      </section>
-      <section>
-        <h2>
-          <code>MediaStream</code> interface
-        </h2>
-        <pre class="idl">
-          partial interface MediaStream {
-              sequence&lt;MediaStreamTrack&gt; getDepthTracks();
-          };
-        
-</pre>
-        <p dfn-for="MediaStream" link-for="MediaStream">
-          The <code><dfn>getDepthTracks</dfn>()</code> method, when invoked,
-          MUST return a sequence of <a data-lt="depth stream track">depth
-          stream tracks</a> in this stream.
-        </p>
-        <p>
-          The <dfn><code>getDepthTracks()</code></dfn> method MUST return a
-          sequence that represents a snapshot of all the
-          <a><code>MediaStreamTrack</code></a> objects in this stream's track
-          set whose <a><code>kind</code></a> is equal to "<code>depth</code>".
-          The conversion from the track set to the sequence is <a>user
-          agent</a> defined and the order does not have to be stable between
-          calls.
-        </p>
-        <p>
-          The <a>MediaStream</a> <a>consumer</a> for the <a>depth-only
-          stream</a> and <a>depth+video stream</a> is <a href=
-          "#the-video-element">the <code>video</code> element</a> [[!HTML]].
-        </p>
-        <div class="note">
-          New <a>consumer</a>s may be added in a future version of this
-          specification.
-        </div>
-        <section class="informative">
-          <h3>
-            Implementation considerations
-          </h3>
-          <p>
-            A <a>video stream track</a> and a <a>depth stream track</a> can be
-            combined into one <a>depth+video stream</a>. The rendering of the
-            two tracks are intended to be synchronized. The resolution of the
-            two tracks are intended to be same. And the coordination of the two
-            tracks are intended to be calibrated. These are not hard
-            requirements, since it might not be possible to synchronize tracks
-            from sources.
-          </p>
-        </section>
-      </section>
-      <section>
-        <h2>
-          <code>MediaStreamTrack</code> interface
-        </h2>
-        <p>
-          The <code><dfn>kind</dfn></code> attribute MUST, on getting, return
-          the string "<code>depth</code>" if the object represents a <a>depth
-          stream track</a>.
-        </p>
-        <p>
-          If a <a>MediaStreamTrack</a> of <a>kind</a> "<code>depth</code>" is
-          <a>muted</a> or <a>disabled</a>, it MUST render black frames, or a
-          zero-information-content equivalent.
-        </p>
-      </section>
-      <section>
-        <h2>
-          <code>MediaDeviceInfo</code> interface
-        </h2>
-        <p>
-          The string "<code>depthinput</code>" is the <a>MediaDeviceKind</a>
-          value for the depth camera input device.
-        </p>
-      </section>
-      <section>
-        <h2>
-          Media provider object
-        </h2>
-        <p>
-          A <a href="https://html.spec.whatwg.org/#media-provider-object">media
-          provider object</a> can represent a <a>depth-only stream</a> (and
-          specifically, not a <a>depth+video stream</a>). The <a>user agent</a>
-          MUST support a <a href=
-          "https://html.spec.whatwg.org/#media-element">media element</a> with
-          an <a href=
-          "https://html.spec.whatwg.org/#assigned-media-provider-object">assigned
-          media provider object</a> that is a <a>depth-only stream</a>, and in
-          particular, the <a href=
-          "https://html.spec.whatwg.org/#dom-media-srcobject"><code>srcObject</code></a>
-          IDL attribute that allows the <a href=
-          "https://html.spec.whatwg.org/#media-element">media element</a> to be
-          assigned a <a href=
-          "https://html.spec.whatwg.org/#media-provider-object">media provider
-          object</a> MUST, on setting and getting, behave as specified in
-          [[!HTML]].
-        </p>
-      </section>
-      <section>
-        <h2>
-          The <code>video</code> element
-        </h2>
-        <p>
-          For a <a href=
-          "https://html.spec.whatwg.org/#the-video-element"><code>video</code>
-          element</a> whose <a href=
-          "https://html.spec.whatwg.org/#assigned-media-provider-object">assigned
-          media provider object</a> is a <a>depth-only stream</a>, the <a>user
-          agent</a> MUST, for each pixel of the <a href=
-          "https://html.spec.whatwg.org/#media-data">media data</a> that is
-          represented by a <a>depth map</a>, <a>convert the depth map value to
-          grayscale</a> prior to when the <code>video</code> element is
-          <a href="https://html.spec.whatwg.org/#potentially-playing">potentially
-          playing</a>.
-        </p>
-        <p>
-          For a <a href=
-          "https://html.spec.whatwg.org/#the-video-element"><code>video</code>
-          element</a> whose <a href=
-          "https://html.spec.whatwg.org/#assigned-media-provider-object">assigned
-          media provider object</a> is a <a>depth+video stream</a>, the <a>user
-          agent</a> MUST act as if all the <a>MediaStreamTrack</a>s of kind
-          "<code>depth</code>" were removed prior to when the
-          <code>video</code> element is <a href=
-          "https://html.spec.whatwg.org/#potentially-playing">potentially
-          playing</a>.
-        </p>
-        <p>
-          The algorithm to <dfn>convert the depth map value to grayscale</dfn>,
+          The data type of a <a>depth map</a> is 16-bit unsigned integer. The
+          algorithm to <dfn>convert the depth map value to grayscale</dfn>,
           given a depth map value <var>d</var>, is as follows:
         </p>
         <ol>
@@ -588,7 +413,7 @@
           <li>Apply the <a>rules to convert using range inverse</a> to
           <var>d</var> to obtain quantized value <var>d<sub>8bit</sub>.</var>
           </li>
-          <li>Return <var>d<sub>8bit</sub></var>.
+          <li>Return <var>d<sub>16bit</sub></var>.
           </li>
         </ol>
         <p>
@@ -694,7 +519,7 @@
           "font-size: 70.7%; vertical-align: -0.219em; padding-right: 0.071em;"><span id="MJXc-Node-45"
           class="mjx-mrow" style=""><span id="MJXc-Node-46" class=
           "mjx-mn"><span class="mjx-char MJXc-TeX-main-R" style=
-          "padding-top: 0.372em; padding-bottom: 0.372em;">8</span></span><span id="MJXc-Node-47"
+          "padding-top: 0.372em; padding-bottom: 0.372em;">16</span></span><span id="MJXc-Node-47"
           class="mjx-mi"><span class="mjx-char MJXc-TeX-math-I" style=
           "padding-top: 0.446em; padding-bottom: 0.298em;">b</span></span><span id="MJXc-Node-48"
           class="mjx-mi"><span class="mjx-char MJXc-TeX-math-I" style=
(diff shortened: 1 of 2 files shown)


bbernhar:
I see that WebGL2 is a good fit for a high precision sink; however, I'm less certain on introducing such a hard requirement into the WD. Could we mark this as future/optional, given the uncertain browser support status?

ds-hwang:
The WebGL2 sample code refers to librealsense GLES3 example I'm pushing now. https://github.com/IntelRealSense/librealsense/pull/168

bbernhar:
How does your example code suggest it's necessary to introduce a hard dependency on WebGL2? Please clarify further.

ds-hwang:
@bbernhar It's because the example uses
1. R16UI texture, which WebGL2 supports
2. vertex texture fetch, which WebGL2 supports

In addition, I'll use vertex texture fetch in chromium implementation also. So I want to enable it on the only device supporting GL context higher than ES3.0.

ds-hwang:
@bbernhar Indeed, WebGL2 is quite hard requirement.
How about supporting texture upload to only WebGL2? Other features can be implemented without WebGL dependency.

bbernhar:
Wouldn't `v1` still allow the canvas to be copied into a texture though? I see WebGL2 makes this more convenient (and efficient) but what are the implications of the spec for browsers that have not yet signaled WebGL2 support [1] @anssiko? 

Equally important, what use-case becomes not possible without requiring WebGL2 support? As mentioned per https://github.com/w3c/mediacapture-depth/issues/110#issuecomment-219703066, the focus is to ship `v1` with at-least UC1 and consider other features (e.g. WebGL2) for UC2+ for `v2`.

[1] http://caniuse.com/#feat=webgl2

ds-hwang:
@bbernhar @anssiko Ok, this PR includes only 16bti change. I'll send WebGL change in another PR.

ds-hwang:
@anssiko In addition, as we use 16bit depth value, I'm wondering if can we use linear range instead of range inverse?

anssiko:
For 16 bit depth, I believe range linear is a better fit since we have more bits to spare.

If we support both the linear and inverse conversion, we need to also expose the conversion method that is in effect through the web-facing API.

How to switch between linear and inverse? In older drafts, we used the range linear option if the source bit depth was 8 bit or less, otherwise range inverse. This was due to the fact that we had to squeeze this through the pipeline that converts the value to 8 bits and back, as you know. The conversion `format` was exposed via `MediaTrackSettings` so web developers could recover d using the correct formula, see:

https://www.w3.org/TR/2016/WD-mediacapture-depth-20160226/#dom-rangeformat
https://www.w3.org/TR/2016/WD-mediacapture-depth-20160226/#dfn-rules-to-convert-using-range-linear

ds-hwang:
@anssiko I see. Let's define linear as default. and introduce range inverse in v2 if needed. what do you think?

index.src.html
```diff
@@ -161,16 +161,15 @@
...
       </p>
       <p>
-        Depth cameras usually produce 16-bit depth values per pixel. However,
-        neither the canvas drawing surface used to draw and manipulate 2D
-        graphics on the web platform nor the <code><a>ImageData</a></code>
-        interface used to represent image data support 16 bits per pixel. To
-        address the issue, this specification defines a conversion into a 8-bit
-        grayscale representation of a <a>depth map</a> for consumption by APIs
-        that are limited to 8 bits per pixel.
+        Depth cameras usually produce 16-bit depth values per pixel. So this
```
anssiko: Editorial suggestions:

s/. So this/, so this/

index.src.html
```diff
@@ -161,16 +161,15 @@
...
+        specification defines a 16-bit grayscale representation of a <a>depth
+        map</a>. There are two ways to access the depth values. First of all,
```
anssiko: s/the depth values/the 16 bit depth values/, also I'd probably make this a bulleted list.

index.src.html
```diff
@@ -161,16 +161,15 @@
...
+        it's a <code>16bit</code> texture in [[WEBGL]] after uploading a
+        <a>depth map</a> to the texture. For instance, <code>RGB</code> and
+        <code>LUMINANCE_ALPHA</code> are <code>16bit</code> texture.
```
anssiko: Better define this in http://rawgit.com/ds-hwang/mediacapture-depth/6dd31ecab245a4674b29eec1049a5c1fb2c923c2/index.src.html#implementation-considerations-1

index.src.html
```diff
@@ -349,6 +348,54 @@
...
+          The data type of a <a>depth map</a> is 16-bit unsigned integer. The
+          algorithm to <dfn>convert the depth map value to grayscale</dfn>,
+          given a depth map value <var>d</var>, is as follows:
+        </p>
+        <ol>
+          <li>Let <var>near</var> be the the <a>near value</a>.
+          </li>
+          <li>Let <var>far</var> be the the <a>far value</a>.
+          </li>
+          <li>Apply the <a>rules to convert using range inverse</a> to
```
anssiko: Should we use range linear now that we use 16 bits?

index.src.html
```diff
@@ -349,6 +348,54 @@
...
+          <var>d</var> to obtain quantized value <var>d<sub>8bit</sub>.</var>
+          </li>
+          <li>Return <var>d<sub>16bit</sub></var>.
+          </li>
+        </ol>
+        <p>
+          The <dfn>rules to convert using range inverse</dfn> are as given in
```
anssiko: Change to range linear.

(Was in an earlier version, see:
https://www.w3.org/TR/2016/WD-mediacapture-depth-20160226/#dfn-rules-to-convert-using-range-linear)

index.src.html
```diff
@@ -349,6 +348,54 @@
...
+        <p>
+          `d_(n) = (far * (d - n ear)) / (d * (far - n ear))`
+        </p>
+        <p>
+          `d_(16bit) = floor(d_(n) * 65535)`
+        </p>
+        <div class="note">
+          <p>
+            The depth measurement <var>d</var> (in meter units) is recovered by
+            solving the <a>rules to convert using range inverse</a> for
```
anssiko: Update to range linear recovery.

index.src.html
```diff
@@ -512,52 +559,12 @@
...
-          <li>Return <var>d<sub>8bit</sub></var>.
-          </li>
-        </ol>
-        <p>
-          The <dfn>rules to convert using range inverse</dfn> are as given in
-          the following formula:
-        </p>
-        <p>
-          `d_(n) = (far * (d - n ear)) / (d * (far - n ear))`
+          the <code>video</code> element draws the depth map at the red channel
```
anssiko: s/the/The/ and `<a>depth map</a>`

anssiko:
Thanks for the PR, added some comments.

index.src.html
bbernhar: Is there a reason why it must render as a red image?

index.src.html
bbernhar: Only if the destination format does not match the internal format, 16-bit depth. Otherwise, conversion is required. If we had a 16-bit sink, there is no need to use a predefined conversion (`RangeLinear` or `RangeInverse`).

index.src.html
bbernhar: I would prefer to separate out the sinks (`<video>` vs WebGL). The WebGL formats mentioned do not fully describe the packing format and would be revisited for consideration for `v2`.

index.src.html
ds-hwang: no any reason. do you have better idea?

index.src.html
ds-hwang: done

index.src.html
ds-hwang: done.

index.src.html
ds-hwang: move to "Implementation considerations"
let me defscribe how to pack format in another PR. It's needed in `v1` because `v1` has to support `WebGL1` as it piggyback `<video>`.

index.src.html
ds-hwang: done.

index.src.html
ds-hwang: Change to range linear.
All camera produces something similar to range linear but not exact. so `RangeLinear` is good idea to standardized output.
@bbernhar WDYT?

ds-hwang:
Thank you for feedback. I resolved all comments.
@anssiko could you review again?

index.src.html
bbernhar: The spec already mentions a "grayscale" representation (that can be accomplished by having 8-bit depth applied to each channel for 32-bit RGB). I believe this largely for illustrative purposes but maintains a similar experience that I saw with a SDK "preview" sample app. Unless if we find a better justification to "go red", I would opt to not modify the default representation.

index.src.html
bbernhar: Thanks for explaining (I would note our assumption in the spec), LGTM.

index.src.html
anssiko: Indeed it would be good to add a new top-level section "Sinks", under which consolidate all the sinks, subsection for each:
- `<video>`
- WebGL (1.0 and 2.0)
- Media Capture Stream with Worker
- (in the future, perhaps `RTCPeerConnection`)

index.src.html
```diff
@@ -512,52 +565,12 @@
...
-          <li>Return <var>d<sub>8bit</sub></var>.
-          </li>
-        </ol>
-        <p>
-          The <dfn>rules to convert using range inverse</dfn> are as given in
-          the following formula:
-        </p>
-        <p>
-          `d_(n) = (far * (d - n ear)) / (d * (far - n ear))`
+          The <code>video</code> element draws the depth map at the red channel
```
anssiko: If we use the red channel, we should probably mention:

`<div class="note">It is an implementation detail how the frames are rendered to the screen. For example, the implementation may use a grayscale representation.</div>`

anssiko:
Thanks for resolving the initial comments, I added couple more comments @ds-hwang.

index.html
```diff
@@ -874,6 +697,250 @@
...
+          class="mjx-mn"><span class="mjx-char MJXc-TeX-main-R" style=
+          "padding-top: 0.372em; padding-bottom: 0.372em;">255</span></span></span><span style="border-bottom: 1.3px solid; top: -0.296em; width: 2.265em;"
+          class="mjx-line"></span></span><span style=
+          "height: 2.225em; vertical-align: -0.778em;" class=
+          "mjx-vsize"></span></span></span></span></span></span></span>
+        </p>
+        <div class="note">
+          It is an implementation detail how the frames are rendered to the
+          screen. For example, the implementation may use a grayscale or red
+          channel representation.
```
ds-hwang: @anssiko browser implementation can choose either gray or red representation

ds-hwang:
@anssiko add the note. now spec don't force red representation. browser implementation can choose either gray or red representation

anssiko:
@ds-hwang I think you did not push the latest change to `ds-hwang:depth_webgl`?

ds-hwang:
@anssiko  yes, latest `ds-hwang:depth_webgl`. I can see the PR is the latest.

anssiko:
Oh yes, my bad. You squashed into the existing commit.

anssiko:
@bbernhar Does this PR now address your concerns, can we merge this?

ds-hwang:
@anssiko I always squashed. Is there another better way?

anssiko:
Squashing in fine, alternatively multiple smaller self-contained commits work fine too.

After getting @bbernhar's confirmation we should be able to merge this PR to update the Editor's Draft. I'd like to get all the Chromium implementation feedback considered before we publishing a new Working Draft snapshot, so please open new issues if/when you spot more issues.

index.html
bbernhar: What would be the benefit of introducing a separate video rendering representation that uses a single-channel format (i.e. red)? Is this only for the display rendering or does it also introduce a separate format for extracting the 8-bit depth from `ImageData`?

anssiko: @bbernhar below are some quick thoughts. @ds-hwang might have some other benefits in mind.
- A single channel (red) would make it rather easy to read "depth pixels" from an `ImageData` object using the video+canvas pipeline:

```
canvas.drawImage(depthVideo, 0, 0, w, h);
var imageData = canvas.getImageData(0, 0, w, h);
var data = imageData.data; // type of Uint8ClampedArray

for (var i = 0; i < data.length; i += 4) {
  var depth = data[i];     // red channel repurposed as the depth pixel value
  // green, blue, and alpha channels are zero
  // var green = data[i + 1];
  // var blue  = data[i + 2];
  // var alpha = data[i + 3];
}
```
- Can reuse [Canvas Pixel ArrayBuffer](https://html.spec.whatwg.org/#canvas-pixel-arraybuffer) as is, just repurpose red as depth. All the existing APIs that deal with Canvas Pixel ArrayBuffer will just work.
- Can improve precision for video+canvas pipeline in the future: [`ImageData`](https://html.spec.whatwg.org/#imagedata)`.data` is of type `Uint8ClampedArray`, could be extended to support other [buffer source types](https://heycam.github.io/webidl/#idl-buffer-source-types), e.g. `Uint16Array`. Currently the video+canvas pipeline is disadvantaged over WebGL, since it is restricted to 8-bit precision per depth pixel.

Alternatively, we could use multiple channels, and define our own packing format (similarly to how it is defined for  `WebGLRenderingContext`, i.e. "the R component represents the lower 8 bit value of 16 bit depth value, the G component represents the upper 8 bit value of 16 bit depth value and the value in B component is not defined."). This would mean web developers need to do unpacking in JavaScript (vs in a shader), probably too slow (I have no evidence through).

The note ("It is an implementation detail how the frames are rendered to the screen. For example, the implementation may use a grayscale or red channel representation.") suggests implementations could render the depth map as they see fit when a depth-only stream is assigned as a media provider object regardless of the underlying `ImageData` representation. Consider:

```
var depthVideo = document.querySelector('#video');
depthVideo.srcObject = depthStream;
video.play();
```

Now that I wrote this comment, it appeared to me we'll still need to update this PR:
- Find a better name for the *convert the depth map value to grayscale" algorithm since it is not really grayscale. Suggestions?
- Note the red component of the Canvas Pixel ArrayBuffer is repurposed for depth.
- Update the recovery steps. Now they assume d<sub>16bit</sub>, but in the canvas+video pipeline we have d<sub>8bit</sub>.
- Nice to have (can be done after this lands): define the multiple sinks we have (<video>, WebGL 1 and 2) into their own section for clarity.

bbernhar: I'm not a big fan of the "define our own packing format" approach for the `video+canvas` sink. I would rather advocate to support for `Uint16Array` type of `ImageData.data` which 1) keeps accessing dexels similar to pixels and 2) requires no conversion. WebGL2 will also be equally capable since it already defines `GL_R16*` texture formats that we could reuse.

To stay comparable with mediacapture-main, we render and capture depth (via `video+canvas`) with the same color format. But why would it matter which color channel we use to store our 8-bit depth value? In grayscale, each dexel can still remain the same width of a pixel and any channel can be used to extract the same 8-bit depth value. There is certainly undesirable duplication repeating the 8-bit dexel in a 32-bit pixel; however, 24-bits gets wasted regardless without a new format.

anssiko:
@bbernhar The PR has been updated. Please let us know if it addresses your comments ([1](https://github.com/w3c/mediacapture-depth/pull/130/files/0d39b4148526af5567f6e6676f3c64177aba7432#r68844511) and [2](https://github.com/w3c/mediacapture-depth/pull/130/files/0d39b4148526af5567f6e6676f3c64177aba7432#r69183042)).

See the updated [`video` element](http://rawgit.com/ds-hwang/mediacapture-depth/64face47ca7c444ff24fa184f572bdfd317dd996/index.html#the-video-element) section and the new [`ImageData` interface](http://rawgit.com/ds-hwang/mediacapture-depth/64face47ca7c444ff24fa184f572bdfd317dd996/index.html#imagedata-interface) section. Note we're currently experimenting how to expose the depth map to the [`WebGLRenderingContext`](http://rawgit.com/ds-hwang/mediacapture-depth/64face47ca7c444ff24fa184f572bdfd317dd996/index.html#webglrenderingcontext-interface), so that section is marked as "work in progress" (aka v2) and is subject to change.

bbernhar:
@anssiko I see **6.7** suggests to split the depth data across two separate channels (or two elements in the `Uint8ClampedArray`). This introduces a new format, _RG88_ for the `video+canvas` sink. I do not see this being intuitive or more efficient. For example, texture filtering algorithms often work with single channel data [1]. When as a WebGL1 texture, additional fragment shader unpacking is required upon sampling (converting RG88 to float16).

Also, the inability to later remove the format from the `video+canvas` sink seems problematic. If our motivation here is to enable a 16-bit precision for `v1`, there are far better alternatives to consider per https://github.com/w3c/mediacapture-depth/pull/130#discussion_r69183042.

Given this, I'd opt to not put much further attention expanding the `video+canvas` sink.

[1] https://en.wikipedia.org/wiki/Bilinear_filtering

ds-hwang:
@bbernhar that's reasonable concern. We'vs extensively discussed about it with chromium engineers. https://bugs.chromium.org/p/chromium/issues/detail?id=624436

Note: R16UI is not filterable. RB8 as well as R16UI needs manual bilinear filter.
R16F requires half-float data, so original data must be converted to half float by CPU.
R16F has one more issue, which doesn't allow texture copy.

ds-hwang:
> I would rather advocate to support for Uint16Array type of ImageData.data which 1) keeps accessing dexels similar to pixels and 2) requires no conversion. WebGL2 will also be equally capable since it already defines GL_R16\* texture formats that we could reuse.

@bbernhar your concern makes sense but `R16` has another big issues.

`Uint16Array` addition requires too big change. e.g. how canvas2d.putImageData(imagedata_u16array, x, y)`will work? how`canvas2d.getImageData() will work?`
It may work like this PR. Moreover, if I'm canvas author, I may be very reluctant this proposal.

On the other hands, `R16UI` is also not filterable. `RG8` requires little more shader code

```
  float getFloatFromRG(vec4 rg) {
    return ((rg.g * 256.0) + rg.r) / 256.0;
  }
  float s1 = getFloatFromRG(texture2D(Texture0, coord1));
```

As `canvas2d` deals it as `RG88`, it is not strange for `WebGL` to handle it as `RG8`.
In addition, web developer can use legacy WebGL code for convenience.

```
texImage2D(context.TEXTURE_2D, 0, context.GL_RGBA, context.GL_RGBA, context.GL_UNSIGNED_BYTE, video_16bit)
```

`R16F` is two critical issue.
1. we have to convert realsense data to half float by CPU
2. `R16F` texture is not copyable

So I think `RG8` is good choice for `canvas2d` as well as `WebGL`.

anssiko:
Thanks @ds-hwang for the explanation, and @bbernhar for your suggestions and comments.

It sounds like the most intuitive approach (`Uint16Array` type of `ImageData.data`) is still considered too disruptive by implementers for the existing `CanvasRenderingContext2D` APIs that [operate on `ImageData` objects](https://html.spec.whatwg.org/multipage/scripting.html#canvasimagedata) (actually, the same API shape was initially proposed in the [First Public Working Draft](https://www.w3.org/TR/2014/WD-mediacapture-depth-20141007/#depthdata-interface) late 2014, but had to be removed due to similar concerns from implementers). That means we'll need to retrofit, and RG8 seems like the best compromise we're aware of to enable 16-bit precision for the 2D rendering context.

Even with RG8 in place, we could in the future extend the `CanvasRenderingContext2D` APIs and e.g. introduce its own `ImageData.depthData` of type `Uint16Array` and specify/patch existing APIs that operate on said objects. But if we want to ship this feature near term, we need to come up with a solution that can be retrofitted.

@bbernhar All that considered, do you have better suggestions how to retrofit 16-bit precision to the 2D rendering context? Unless there are any better alternatives, I'd land this PR to update the ED, since it bakes in other clarification as well.

bbernhar:
> On the other hands, R16UI is also not filterable. RG8 requires little more shader code

I'm OK with `R16UI` requiring "manual" texture filtering in order to enable 16-bit precision in WebGL (the single channel keeps this straightforward). I feel like it is a separate issue altogether on how to best enable filterable depth textures (via `R16F` + _oes_texture_half_float_ or something like _os_texture_half_uint_); where knowing the specific Depth + WebGL use-case matters. The `R16UI` format resembles the actual native format (i.e. `D16`) and remains forward compatible.

> As canvas2d deals it as RG88, it is not strange for WebGL to handle it as RG8.

Rather than overload the meaning of `ImageData.data` (being a `Uint8Array` or `Uint16Array` type) - would it not be possible to similarly have a new `DepthData` object (e.g. 16-bit "view" of a canvas buffer)? This would be more align with the native SDKs which separate frame types: `ColorFrame` & `DepthFrame` [1]. This could be a worthy design note to follow.

[1] https://msdn.microsoft.com/en-us/library/windowspreview.kinect.depthframe.aspx

anssiko:
@bbernhar Your suggestion for a new `DepthData` object sounds like a good v2 feature. Could you open a new issue for that so we could incubate the proposal further.

Meanwhile, I'll merge this PR since we believe it is the best retrofit solution for 16-bit precision.

As usual, we will continue improve the specification in subsequent edits.

PR #139, opened 2016-10-31
Fix #137: Define unknown depth map value

### LINKED ISSUE #137: Need an invalid value
This bug is important to fix before this goes for wider review. 

Most depth sensors may not product a depth for every pixel. It is important that when they can no detect what the depth value is at a give location, they can report that there is no reading vs reporting a bad reading. We need to define a way a value that indicates "no depth detected". Typically this is zero. Once we define this value, then the near / far distance calculations formulas need to be adjusted.

anssiko:


fluffy:
This look right but do you need more? I would assume this might change some of the formulas in the draft

anssiko:
@huningxin to review if changes are needed.

huningxin:
Thanks for this PR.

However, IMO, this PR only changes the depth-to-grayscale of "The algorithm to convert the depth map value to grayscale". I suppose we need:
1. define the 0 is the invalid value of depth value in depth map.
2. define the invalid depth value case of the depth-to-grayscale conversion (as this PR does).
3. define the invalid depth value case of the grayscale-to-depth conversion.

anssiko:
@huningxin:

>1. define the 0 is the invalid value of depth value in depth map.

I attempted to address this in https://github.com/w3c/mediacapture-depth/pull/139/commits/fa43616ed0452062813629e094731f37803725f4.

>3. define the invalid depth value case of the grayscale-to-depth conversion.

Can you clarify what you mean with the grayscale-to-depth conversion? We currently only define depth-to-grayscale for the purpose of rendering depth stream in a video element: https://w3c.github.io/mediacapture-depth/index.src.html#dfn-convert-the-depth-map-value-to-grayscale

I've been away from this spec for some time, so I'm probably missing something obvious :-)

huningxin:
> Can you clarify what you mean with the grayscale-to-depth conversion?

I mean the `NOTE` section about "The depth measurement d (in meter units) is recovered by ......".

anssiko:
>I mean the NOTE section about "The depth measurement d (in meter units) is recovered by ......".

Given an invalid depth map value `d_(16bit) = 0`, the grayscale-to-depth normalize step returns 0:

`d_(n) = d_(16bit) / 65535 = 0`

Substituting `d_(n)` with 0 in the the conversion step:

`d = (d_(n) * (far - near)) + near` reduces to `d = near`

To check the operation is invertible in this special case, we substitute `d` with `near` in depth-to-grayscale conversion:

`d_(n) = (d - near) / (far - near)` reduces to `d_(n) = 0`

And so does `d_(16bit)`:

`d_(16bit) = floor(d_(n) * 65535) = 0`

Q.E.D.

What type of additional text you'd prefer us to add to the NOTE section for this special case?

huningxin:
I think we need tell the case that if the `d_(16bit)` is `0` (invalid), then the `d_(n)` is also 0 (the invalid value).

WDYT?

anssiko:
@huningxin I added https://github.com/w3c/mediacapture-depth/pull/139/commits/8f753dcde55389f50ba2548b0908fab7a1b85f4b to make that case obvious. PTAL. Do you think we have now addressed this issue and can resolve and land?

huningxin:
Sorry about the confusion, I made a mistake in https://github.com/w3c/mediacapture-depth/pull/139#issuecomment-271296132 by suing `d_(n)` where it should be `d`.

So it would be "If the `d_(16bit)` is 0, then the `d` is the invalid depth map value."

anssiko:
@huningxin Updated, does https://github.com/w3c/mediacapture-depth/pull/139/commits/c91cb79eb138a0cc36c42468b610e2ef807ddb63 look better? Moved this into recovery steps proper.

huningxin:
LGTM. Thanks for the PR!

PR #141, opened 2016-10-31
Issue 134 drop redundant xdm reference

anssiko:


anssiko:
No concerns raised, merged.

PR #144, opened 2016-11-15
Fix #110: Intrinsics and extrinsics

### LINKED ISSUE #110: Accessing Camera Calibrations

### LINKED ISSUE #135: Focal length seems under defined
I don't understand how to use focal length without understanding the size of the sensor in mm. If the focal length was 50 mm, what would that mean ? what would an application use it for ?

astojilj:
Fix #110: Intrinsics and extrinsics
    
Principal point, distortion coefficients, depth to video transformation.
Formulas for deprojection, transform and projection.
TODO. shader example code.    
Fix #110: Accessing Camera Calibrations

### DIFF
--- a/build.js
+++ b/build.js
@@ -62,15 +62,14 @@ mjAPI.typeset({
   var re2 = /.mjx-span {display: span}\n/gi;
   
   // remove the MathJax.js <script>
-  var re3 = /\<script src="https:\/\/cdn.mathjax.org\/mathjax\/latest\/MathJax.js\?config=AM_CHTML"\>/gi;
-
-  // remove the MathJax @font-face CSS at-rules
-  // var re4 = /@font-face {font-family: MJX(.*?)}\n/gi;
+  var re3 = /\<script src="https:\/\/cdn.mathjax.org\/mathjax\/latest\/MathJax.js\?config=AM_CHTML"\>\n<\/script>/gi;
 
   var HTML = HTML.replace(re1, "").replace(re2, "").replace(re3, "");
     
   // tidy the generated output and write to index.html
-  var HTML = tidy(HTML, tidyConfig);
+  // Breaks (mostly indention) complex MathJax formulas. Disabled for now. 
+  // var HTML = tidy(HTML, tidyConfig);
+
   fs.writeFileSync('index.html', HTML, 'utf8');
   
   process.exit();
--- a/index.src.html
+++ b/index.src.html
@@ -265,6 +265,11 @@
         "!WEBIDL#idl-Uint16Array">Uint16Array</dfn></code> buffer source types
         are defined in [[WEBIDL]].
       </p>
+      <p>
+        The meaning of dictionary member being <dfn data-cite=
+        "!WEBIDL#dfn-present">present</dfn> or <dfn data-cite=
+        "!WEBIDL#dfn-present">not present</dfn> is defined in [[WEBIDL]].
+      </p>
     </section>
     <section>
       <h2>
@@ -326,13 +331,50 @@
         </p>
         <p>
           A <a>depth map</a> has an associated <dfn>horizontal focal
-          length</dfn> which is a double. It represents the horizontal focal
-          length of the depth camera, in pixels.
+          length</dfn> which is a double. It represents the horizontal
+          <dfn>focal length</dfn> of the depth camera, in pixels.
         </p>
         <p>
           A <a>depth map</a> has an associated <dfn>vertical focal length</dfn>
           which is a double. It represents the vertical focal length of the
           depth camera, in pixels.
+        </p>
+        <p>
+          A <a>depth map</a> has an associated <dfn>principal point</dfn>,
+          specified by <dfn>principal point x</dfn> and <dfn>principal point
+          y</dfn> coordinates which are double. It is a concept defined in the
+          pinhole camera model; a projection of perspective center to the image
+          plane.
+        </p>
+        <p>
+          A <a>depth map</a> has an associated <dfn>transformation from depth
+          to video</dfn>, which is a <dfn>transformation matrix</dfn>
+          represented by a <a>DepthToVideoTransform</a> dictionary. It is used
+          to translate position in depth camera 3D coordinate system to RGB
+          video stream's camera (identified by <dfn>videoDeviceId</dfn>) 3D
+          coordinate system. After projecting depth 2D pixel coordinates to 3D
+          space, we use this matrix to transform depth camera 3D space
+          coordinates to RGB video camera 3D space.
+        </p>
+        <p>
+          Both depth and color cameras usually introduce significant distortion
+          caused by the camera and lens. While in some cases, the effects are
+          not noticeable, these distortions cause errors in image analysis. To
+          map <a>depth map</a> pixel values to corresponding color video track
+          pixels, we use two <a>DistortionCoefficients</a> dictionaries:
+          <dfn>deprojection distortion coefficients</dfn> and <dfn>projection
+          distortion coefficients</dfn>.
+        </p>
+        <p>
+          <a>Deprojection distortion coefficients</a> are used for compensating
+          camera distortion when deprojecting 2D pixel coordinates to 3D space
+          coordinates. <a>Projection distortion coefficients</a> are used in
+          the opposite case, when projecting camera 3D space points to pixels.
+          One track doesn't have both of the coefficients specified. The most
+          common scenario is that the depth track has <a>deprojection
+          distortion coefficients</a> or that the color video track has
+          <a>projection distortion coefficients</a>. For the details, see
+          <a>algorithm to map depth pixels to color pixels</a>.
         </p>
         <p>
           The data type of a <a>depth map</a> is 16-bit unsigned integer. The
@@ -407,6 +449,11 @@
               boolean depthFar = true;
               boolean focalLengthX = true;
               boolean focalLengthY = true;
+              boolean principalPointX = true;
+              boolean principalPointY = true;
+              boolean deprojectionDistortionCoefficients = false;
+              boolean projectionDistortionCoefficients = false;
+              boolean depthToVideoTransform = false;
           };
 </pre>
         <section>
@@ -468,6 +515,61 @@
               "#def-constraint-focalLengthY">focalLengthY</a></code> for
               details.
             </dd>
+            <dt>
+              <dfn data-dfn-for=
+              "MediaTrackSupportedConstraints"><code>principalPointX</code></dfn>
+              of type <span class="idlMemberType"><a>boolean</a></span>,
+              defaulting to <code>true</code>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointX">principalPointX</a></code> for
+              details.
+            </dd>
+            <dt>
+              <dfn data-dfn-for=
+              "MediaTrackSupportedConstraints"><code>principalPointY</code></dfn>
+              of type <span class="idlMemberType"><a>boolean</a></span>,
+              defaulting to <code>true</code>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointY">principalPointY</a></code> for
+              details.
+            </dd>
+            <dt>
+              <dfn data-dfn-for=
+              "MediaTrackSupportedConstraints"><code>deprojectionDistortionCoefficients</code></dfn>
+              of type <span class="idlMemberType"><a>boolean</a></span> ,
+              defaulting to <code>false</code>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-deprojectionDistortionCoefficients">deprojectionDistortionCoefficients</a></code>
+              for details.
+            </dd>
+            <dt>
+              <dfn data-dfn-for=
+              "MediaTrackSupportedConstraints"><code>projectionDistortionCoefficients</code></dfn>
+              of type <span class="idlMemberType"><a>boolean</a></span> ,
+              defaulting to <code>false</code>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-projectionDistortionCoefficients">projectionDistortionCoefficients</a></code>
+              for details.
+            </dd>
+            <dt>
+              <dfn data-dfn-for=
+              "MediaTrackSupportedConstraints"><code>depthToVideoTransform</code></dfn>
+              of type <span class="idlMemberType"><a>boolean</a></span> ,
+              defaulting to <code>false</code>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-depthToVideoTransform">depthToVideoTransform</a></code>
+              for details.
+            </dd>
           </dl>
         </section>
       </section>
@@ -482,6 +584,11 @@
               (double or DoubleRange) depthFar;
               (double or DoubleRange) focalLengthX;
               (double or DoubleRange) focalLengthY;
+              (double or DoubleRange) principalPointX;
+              (double or DoubleRange) principalPointY;
+              boolean deprojectionDistortionCoefficients;
+              boolean projectionDistortionCoefficients;
+              boolean depthToVideoTransform;
           };
 </pre>
         <section>
@@ -532,6 +639,54 @@
               "#def-constraint-focalLengthY">focalLengthY</a></code> for
               details.
             </dd>
+            <dt>
+              <dfn><code>principalPointX</code></dfn> of type <span class=
+              "idlMemberType"><a>(double or DoubleRange)</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointX">principalPointX</a></code> for
+              details.
+            </dd>
+            <dt>
+              <dfn><code>principalPointY</code></dfn> of type <span class=
+              "idlMemberType"><a>(double or DoubleRange)</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointY">principalPointY</a></code> for
+              details.
+            </dd>
+            <dt>
+              <dfn><code>deprojectionDistortionCoefficients</code></dfn> of
+              type <span class="idlMemberType"><a>boolean</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-deprojectionDistortionCoefficients">deprojectionDistortionCoefficients</a></code>
+              for details. Returns false if the source cannot provide this
+              information.
+            </dd>
+            <dt>
+              <dfn><code>projectionDistortionCoefficients</code></dfn> of type
+              <span class="idlMemberType"><a>boolean</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-projectionDistortionCoefficients">projectionDistortionCoefficients</a></code>
+              for details. Returns false if the source cannot provide this
+              information.
+            </dd>
+            <dt>
+              <dfn><code>depthToVideoTransform</code></dfn> of type
+              <span class="idlMemberType"><a>boolean</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-depthToVideoTransform">depthToVideoTransform</a></code>
+              for details. Returns false if the source cannot provide this
+              information.
+            </dd>
           </dl>
         </section>
       </section>
@@ -546,6 +701,11 @@
               ConstrainDouble depthFar;
               ConstrainDouble focalLengthX;
               ConstrainDouble focalLengthY;
+              ConstrainDouble principalPointX;
+              ConstrainDouble principalPointY;
+              ConstrainBoolean deprojectionDistortionCoefficients;
+              ConstrainBoolean projectionDistortionCoefficients;
+              ConstrainBoolean depthToVideoTransform;
           };
 </pre>
         <section>
@@ -596,6 +756,51 @@
               "#def-constraint-focalLengthY">focalLengthY</a></code> for
               details.
             </dd>
+            <dt>
+              <dfn><code>principalPointX</code></dfn> of type <span class=
+              "idlMemberType"><a>ConstrainDouble</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointX">principalPointX</a></code> for
+              details.
+            </dd>
+            <dt>
+              <dfn><code>principalPointY</code></dfn> of type <span class=
+              "idlMemberType"><a>ConstrainDouble</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointY">principalPointY</a></code> for
+              details.
+            </dd>
+            <dt>
+              <dfn><code>deprojectionDistortionCoefficients</code></dfn> of
+              type <span class="idlMemberType"><a>ConstrainBoolean</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-deprojectionDistortionCoefficients">deprojectionDistortionCoefficients</a></code>
+              for details.
+            </dd>
+            <dt>
+              <dfn><code>projectionDistortionCoefficients</code></dfn> of type
+              <span class="idlMemberType"><a>ConstrainBoolean</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-projectionDistortionCoefficients">projectionDistortionCoefficients</a></code>
+              for details.
+            </dd>
+            <dt>
+              <dfn><code>depthToVideoTransform</code></dfn> of type
+              <span class="idlMemberType"><a>ConstrainBoolean</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-depthToVideoTransform">depthToVideoTransform</a></code>
+              for details.
+            </dd>
           </dl>
         </section>
       </section>
@@ -610,6 +815,24 @@
               double              depthFar;
               double              focalLengthX;
               double              focalLengthY;
+              double              principalPointX;
+              double              principalPointY;
+              DistortionCoefficients deprojectionDistortionCoefficients;
+              DistortionCoefficients projectionDistortionCoefficients;
+              DepthToVideoTransform depthToVideoTransform;
+          };
+
+          dictionary DistortionCoefficients {
+              double              k1;
+              double              k2;
+              double              p1;
+              double              p2;
+              double              k3;
+          };
+
+          dictionary DepthToVideoTransform {
+            Float32Array       transformationMatrix;
+            DOMString          videoDeviceId;
           };
 </pre>
         <section>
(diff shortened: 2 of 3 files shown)


astojilj:
@huningxin, @robman, @bbernhar, @fluffy, PTAL.
Thanks.

astojilj:
@huningxin, I know you are busy. Some feedback would be appreciated as I'm starting work on writing formulas for deprojection etc... Thanks.

huningxin:
@astojilj , Thanks for your PR! It's my bad that I didn't comment here. 

But I understand that we still have some opens in #110 and #135.  

My suggestion would be only including the changes with consensus in this PR and leave the opens in following PRs. WDYT?

astojilj:
@huningxin, @robman, @anssiko 
>But I understand that we still have some opens in #110 and #135.

#135:
I don't see there is anything open in #135. All the comments and reporter confirm that this needs to be fixed and the PR here is a fix for it. in [comment here](https://github.com/w3c/mediacapture-depth/issues/135#issuecomment-260585403) @robman expresses agreement with my approach coded here in this PR.

#110,
The only open there is @robman's [proposal](https://github.com/w3c/mediacapture-depth/issues/110#issuecomment-260587927) about which part of this PR should be v1 and which should be v2 scope. You also raised the [question](https://github.com/w3c/mediacapture-depth/issues/110#issuecomment-260875248) here about why to attempt to provide incomplete data only.

I don't think that PR should matter about v1/v2 scope split. Or in PR only we should put content that all the participants in issue page agree should be v1  and delete all v2 scope content?

huningxin:
Then how about separate this PR into one for #135 and another one for #110. It would be more clear to me. Thanks!

astojilj:
Yes, let me do so. Thanks.

astojilj:
PR for #135 is split to https://github.com/w3c/mediacapture-depth/pull/146.

anssiko:
Updated the title, since #135 was already fixed with https://github.com/w3c/mediacapture-depth/commit/e50bc08af1bc6e520d87dd66d6842191b322ac6c

astojilj:
Updated. PTAL. 
Travis CI fails on acceptable "Warning: trimming empty <span>" only.

astojilj:
After peer review with Anssi:
- need to generalize distortion model and coefficients (probably remove the model name)
- provide generic formula for applying distortion coefficients.

astojilj:
Updated the pull request by:
1) Removed device specific distortion models
2) added full formulas for depth deprojection, translating from depth to color coordinates and color projection with the optional distortion handling formular on both deptojection and projection.
3) transformation matrix was modeled using translation and quaternion for rotation - replaced them by simple 16 element array in the format suitable to be directly used in matrix x vector multiplication and with WebGL's uniformMatrix4fv.

Example code for color deprojection and projection I'll add in separate request - don't want to grow this one further.

index.src.html
```diff
@@ -345,15 +345,232 @@
...
         <p>
+          A <a>depth map</a> has an associated <dfn>principal point</dfn>,
+          specified by <dfn>principal point x</dfn> and <dfn>principal point
+          y</dfn> coordinates which are double. It is a concept defined in the
+          pinhole camera model; a projection of perspective center to the image
+          plane.
+        </p>
+        <p>
+          A <a>depth map</a> has an associated <dfn>transformation from depth
+          to video</dfn>, which is <a>DepthToVideoTransform</a> dictionary. It
```
anssiko: `which is a <dfn>transformation matrix</dfn> represented by a <a>DepthToVideoTransform</a> dictionary`

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+        </p>
+        <p>
+          Deprojection coefficients are used for compensating camera distortion
+          when deprojecting 2D pixel coordinates to 3D space coordinates.
+          Projection coefficients are used in the opposite case, when
+          projecting camera 3D space points to pixels. One track doesn't have
+          both of the coefficients specified. The most common scenario is that
+          the depth track has deprojection coefficients or that the color video
+          track has projection coefficients.
+        </p>
```
anssiko: L390 onwards could be in its own section, for example (adjust titles so they make sense):

* Coordinate transformation from depth to video
  * Deproject depth map values to points in depth 3D space
  * Transform 3D points from depth camera 3D space to color camera 3D space
  * Project from color camera 3D space to color frame 2D pixels
* Conversion from the depth map values to grayscale
  * (this would contain the current _convert the depth map value to grayscale_ and _rules to convert using range linear_ algorithms)

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+        <p>
+          With this case in mind, the algorithm for mapping depth pixels to
```
anssiko: `The <dfn>algorithm to map depth pixels to color pixels</dfn> is as follows:`

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+          color pixels is as follows:
+        </p>
+        <ol>
+          <li>Deproject <a>depth map</a> values to points in depth 3D space.
```
anssiko: `When the user agent is to <dfn>deproject depth map values</dfn> to points in depth 3D space, it must run the following algorithm:`

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+              Let <var>dx</var> and <var>dy</var> be 2D coordinates, in pixels,
+              of a pixel in <a>depth map</a>.
+            </p><var></var>
+            <p>
+              Let <var>dz</var> be <a>depth map value</a> of the same pixel in
+              the <a>depth map</a>.
+            </p>
+            <p>
+              Let <var>fx</var> and <var>fy</var> be <a>depth map</a>'s
+              horizontal and vertical <a>focal length</a>s.
```
anssiko: `<a>horizontal focal length</a> and <a>vertical focal length</a> respectively.

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+            </p>
+            <p>
+              `px = (dx - cx) / (fx)`
+            </p>
+            <p>
+              `py = (dy - cy) / (fy)`
+            </p>
+            <ul>
+              <li>If <a>depth map</a>'s <a>deprojection distortion
+              coefficients</a> are not available,
```
anssiko: Must define what "not available" means, `DistortionCoefficients` is set to false?

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+                <p>
+                  `Yd = dz * px`
+                </p>
+                <p>
+                  `Zd = dz`
+                </p>
+              </li>
+              <li>If <a>depth map</a>'s <a>deprojection distortion
+              coefficients</a> <a>k1</a>, <a>k2</a>, <a>k3</a>, <a>p1</a> and
+              <a>p2</a> are available, with a note that some of those could be
```
anssiko: s/are available/are not omitted/

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+              zero,
+                <p>
+                  3D coordinates (Xd, Yd, Zd) in depth camera space are
```
anssiko: `When the user agent is to <dfn>calculate 3D coordinates in depth camera space</dfn> (Xd, Yd, Zd), it must run the following algorithm:`

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+                <p>
+                  `Zd = dz`
+                </p>
+              </li>
+            </ul>
+          </li>
+          <li>Transform 3D points from depth camera 3D space to color camera 3D
+          space.
+            <p>
+              The 3D point (Xd, Yd, Zd) calculated in previous step is in depth
```
anssiko: Link to the previous section rather than say "previous step":

To cross-link easily, you can say `<a>foo</a>` to link to `<dfn>foo</dfn>` globally in the document.

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+          <li>Finally, project from color camera 3D space to color frame 2D
+          pixels.
+            <p>
+              With this step, we conclude mapping of a <a>depth map</a> pixel
+              to a pixel value in color video frame.
+            </p>
+            <p>
+              For projection we use color track's <a>projection distortion
+              coefficients</a> <a>k1</a>, <a>k2</a>, <a>k3</a>, <a>p1</a> and
+              <a>p2</a> if available. The color track that corresponds to the
```
anssiko: s/if available/if not omitted/

What happens if they are omitted? Define failure path.

index.src.html
```diff
@@ -345,15 +345,232 @@
...
+                  `x = (px_c + 2 * p1 * px_c * py_c + p2 * (r2_c + 2 * px_c^2))
+                  * fx_c + cx_c`
+                </p>
+                <p>
+                  `y = (py_c + 2 * p2 * px_c * py_c + p1 * (r2_c + 2 * py_c^2))
+                  * fy_c + cy_c`
+                </p>
+              </li>
+              <li>If color track's <a>projection distortion coefficients</a>
+              are not available
```
anssiko: are omitted

anssiko:
Thanks! I added some review comments.

In addition, you'll need to patch the build.js as follows to make MathJax happy with the complex formulas:

https://gist.github.com/anssiko/fd03fb67eb7b29dd1abecbac0ec43536

index.src.html
astojilj: Thanks. Done.

index.src.html
astojilj: _Coordinate transformation from depth to video_ is separated now.
_Conversion from the depth map values to grayscale_ I'm doing in separate patch as it requires content changes.

index.src.html
astojilj: done.

index.src.html
astojilj: Application or user agent. rephrased.

index.src.html
astojilj: done.

index.src.html
astojilj: done.

index.src.html
astojilj: It is defined under - rewrote this part so there is no confusion. "available" should probably be "defined".

index.src.html
astojilj: done.

index.src.html
astojilj: defined?

index.src.html
astojilj: i use omitted for undefined and defined instead of available now.

in JS, "defined" would be  `if(track.getSettings().deprojectionDistortionCoefficients)   ` - added this note as inline example.

astojilj (commented):
PR updated.
Thanks.

anssiko:
Regarding checking the presence of dictionary members, we should actually check if the member is present, and refer to https://heycam.github.io/webidl/#dfn-present like this:

`"If the <a>foobar</a> member is <a>present</a> in <a>baz dictionary</a>"`

(Additionally, add appropriate `<dfn>present</dfn>` to Dependencies.

Related issue: https://github.com/heycam/webidl/issues/124

astojilj:
<a>present</a> and <a>not present</a> in <a href="#mediatracksettings-dictionary">MediaTrackSettings</a> dictionary in updated PR.

index.src.html
```diff
@@ -326,15 +331,51 @@
...
+          plane.
+        </p>
+        <p>
+          A <a>depth map</a> has an associated <dfn>transformation from depth
+          to video</dfn>, which is a <dfn>transformation matrix</dfn>
+          represented by a <a>DepthToVideoTransform</a> dictionary. It is used
+          to translate position in depth camera 3D coordinate system to RGB
+          video stream's camera (identified by <dfn>videoDeviceId</dfn>) 3D
+          coordinate system. After projecting depth 2D pixel coordinates to 3D
+          space, we use this to translate depth camera 3D space coordinates to
```
anssiko: we use this projection?

index.src.html
```diff
@@ -326,15 +331,51 @@
...
+          RGB video camera 3D space.
+        </p>
+        <p>
+          Both depth and color cameras usually introduce significant distortion
+          caused by the camera and lens. While in some cases, the effects are
+          not noticeable, these distortions cause errors in image analysis. To
+          map <a>depth map</a> pixel values to a corresponding color video
```
anssiko: to corresponding

index.src.html
```diff
@@ -326,15 +331,51 @@
...
+          track pixels, we use two <a>DistortionCoefficients</a> dictionaries:
+          <dfn>deprojection distortion coefficients</dfn> and <dfn>projection
+          distortion coefficients</dfn>.
+        </p>
+        <p>
+          Deprojection distortion coefficients are used for compensating camera
```
anssiko: `<a>Deprojection distortion coefficients</a>`

index.src.html
```diff
@@ -326,15 +331,51 @@
...
+          distortion when deprojecting 2D pixel coordinates to 3D space
+          coordinates. Projection distortion coefficients are used in the
```
anssiko: `<a>Projection distortion coefficients</a>`

index.src.html
```diff
@@ -326,15 +331,51 @@
...
+          opposite case, when projecting camera 3D space points to pixels. One
+          track doesn't have both of the coefficients specified. The most
+          common scenario is that the depth track has deprojection coefficients
```
anssiko: `<a>deprojection distortion coefficients</a>`

index.src.html
```diff
@@ -326,15 +331,51 @@
...
+          or that the color video track has projection coefficients. For the
```
anssiko: `<a>projection distortion coefficients</a>`

index.src.html
```diff
@@ -394,6 +435,233 @@
     </section>
     <section>
       <h2>
+        Synchronizing depth and color video rendering
+      </h2>
+      <p>
+        When rendering, we want to position a color from color video frame to
```
anssiko: a color value?

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+        corresponding <a>depth map value</a> or 3D point in space defined by
+        <a>depth map value</a>. We use <a>deprojection distortion
+        coefficients</a> to compensate camera distortion when deprojecting 2D
+        pixel coordinates to 3D space coordinates and <a>projection distortion
+        coefficients</a> in the opposite case, when projecting camera 3D space
+        points to pixels. The algorithm here intents to cover all of the cases.
```
anssiko: `The <a>algorithm to map...</a> and <a>algorithm to deproject...</a> are defined in this specifications.`

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+        Most commonly, either the depth track has deprojection or the color
```
anssiko: We already say above that "The most common scenario is that the depth track has deprojection coefficients or that the color video track has projection coefficients" so we do not need to repeat.

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          <a>Project</a> from color camera 3D space to color frame 2D pixels.
+        </li>
+      </ol>
+      <p></p>
+      <section>
+        <h2>
+          <dfn>Deproject</dfn> to depth 3D space
+        </h2>
+        <p>
+          To <a>deproject</a> <a>depth map value</a> to point in depth camera
```
anssiko: `The <dfn>algorithm to deproject depth map value</dfn> to point in depth camera is as follows:`

anssiko:
```
To <dfn>deproject</dfn> a depth map value to point in depth camera 3D space, the user agent must run the <a>algorithm to deproject depth map value</a>.

(Remove dfns from headers.)

The <dfn>algorithm to deproject depth map value</dfn> to point in depth camera is as follows:
```

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          3D space, application or user agent must run following algorithm:
+        </p>
+        <p>
```
anssiko: Make this and other algorithms numbered lists, nested at ifs:

```
<ol> 
  <li>...</li> 
  <li>If ...
    <ol> 
      <li>inner ...</li> 
      <li>...</li> 
    </ol>
  </li>
  <li>...</li> 
</ol> 
```

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          In code, this condition would be like
+          <code>if(track.getSettings().deprojectionDistortionCoefficients)</code>
+        </p>
+      </section>
+      <section>
+        <h2>
+          <dfn>Transform</dfn> from depth to color 3D space
+        </h2>
+        <p>
+          The 3D point (Xd, Yd, Zd), that is a result of <a>project</a> depth
```
anssiko: `To <dfn>transform</dfn> xxxxxx, the user agent must run the <a>algorithm to transform yyyy</a>.`

`The <dfn>algorithm to transform yyyy</dfn> is as follows:`

(Remove dfn from the header.)

This allows us to decouple the "transform" concept from its algorithm, so we can link to them separately as needed elsewhere in the spec.

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          value to 3D point step, is in depth camera 3D coordinate system. To
+          get the coordinates of the same point in space, but in color camera
+          3D coordinate system, we use matrix multiplication of
+          <a>transformation from depth to video</a> matrix by the (Xd, Yd, Zd)
+          3D point vector.
+        </p>
+        <p>
```
anssiko: Make this a numbered list.

index.src.html
anssiko: Alternative, more concise way to say this (if you won't need to link to the algorithm directly elsewhere in the spec):
`To <dfn>deproject</dfn> a depth map value to point in depth camera 3D space, the user agent must run the following steps:`

index.src.html
anssiko: (See related comment above for more concise way to spec this.)

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          `((Xc), (Yc), (Zc)) = ([M] xx ((Xd), (Yd), (Zd), (1))).xyz`
+        </p>
+      </section>
+      <section>
+        <h2>
+          <dfn>Project</dfn> from color 3D to pixel
+        </h2>
+        <p>
+          With this step, we conclude mapping of a <a>depth map value</a>
```
anssiko: Here, use the same style as above:

`To <dfn>project</dfn> ...`

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          corresponding color track using <a>depth map</a>'s
+          <a>depthToVideoTransform</a> field <a>videoDeviceId</a> - it
+          represents the target color video deviceID that should be used as a
+          constraint with [[!GETUSERMEDIA]] call to get the corresponding color
+          video stream track. After that, we use color track
+          <a>getSettings()</a> to access <a href=
+          "#mediatracksettings-dictionary">MediaTrackSettings</a>.
+        </p>
+        <p>
+          Let `fx_c` and `fy_c` be color track's horizontal and vertical
```
anssiko: `<a>color stream track</a>'s <a>horizontal focal length</a> and <a>vertical focal length</a> respectively.`

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+      </section>
+      <section>
+        <h2>
+          <dfn>Project</dfn> from color 3D to pixel
+        </h2>
+        <p>
+          With this step, we conclude mapping of a <a>depth map value</a>
+          coordinate to a pixel value in color video frame.
+        </p>
+        <p>
```
anssiko: It sounds like this para is informative content? If so, use:
`<p class="note">`

index.src.html
```diff
@@ -394,6 +435,233 @@
...
+          "#mediatracksettings-dictionary">MediaTrackSettings</a>. We get the
+          corresponding color track using <a>depth map</a>'s
+          <a>depthToVideoTransform</a> field <a>videoDeviceId</a> - it
+          represents the target color video deviceID that should be used as a
+          constraint with [[!GETUSERMEDIA]] call to get the corresponding color
+          video stream track. After that, we use color track
+          <a>getSettings()</a> to access <a href=
+          "#mediatracksettings-dictionary">MediaTrackSettings</a>.
+        </p>
+        <p>
```
anssiko: `<p>To <dfn>project</dfn> xxxxx, the user agent must run the following steps:</p>`

And make the steps a numbered list.

anssiko:
Discussed with @astojilj and concluded "Synchronizing depth and color video rendering" section will be made non-normative guidance to web developers. Thus some of my earlier comments are not valid, since we should not use RFC terms in an informative section.

index.src.html
astojilj: ok, agreed to leave it as it is.

index.src.html
astojilj: done.

index.src.html
astojilj: done.

index.src.html
astojilj: "this matrix to transform"

index.src.html
astojilj: done

index.src.html
astojilj: done

index.src.html
astojilj: done

index.src.html
astojilj: done

index.src.html
astojilj: done

index.src.html
astojilj: redundant sentence, removed.

index.src.html
astojilj: done.

index.src.html
astojilj: acknowledged. ignored as this goes to example section.

index.src.html
astojilj: removed as not very informative.

astojilj:
PR updated. Section moved just before examples, Example vertex shader code added and links to it added around.
@ds-hwang Kudos for the [code](https://github.com/IntelRealSense/librealsense/pull/168/files#diff-eb62c5bc83e467cd92828ec977b8a77bR439) - I used it many times last week.

astojilj:
@anssiko PTAL

build.js
```diff
@@ -62,15 +62,19 @@ mjAPI.typeset({
   var re2 = /.mjx-span {display: span}\n/gi;
   
   // remove the MathJax.js <script>
-  var re3 = /\<script src="https:\/\/cdn.mathjax.org\/mathjax\/latest\/MathJax.js\?config=AM_CHTML"\>/gi;
+  var re3 = /\<script src="https:\/\/cdn.mathjax.org\/mathjax\/latest\/MathJax.js\?config=AM_CHTML"\>\n<\/script>/gi;
 
   // remove the MathJax @font-face CSS at-rules
   // var re4 = /@font-face {font-family: MJX(.*?)}\n/gi;
 
+  var re4 = /.mjx-char {display: block; white-space: pre}\n/gi;
```
anssiko: re4 seems redundant, is not used in L72.

index.src.html
```diff
@@ -660,8 +883,104 @@
...
+              "idlMemberType"><a>double</a></span>
+            </dt>
+            <dd>
+              See <code><a href=
+              "#def-constraint-principalPointY">principalPointY</a></code> for
+              details.
+            </dd>
+            <dt>
+              The <dfn><code>deprojectionDistortionCoefficients</code></dfn>
+              dictionary member represents the media track's (usually the
```
anssiko: Ambiguos. Suggestion: "represents depth map's distortion...".

index.src.html
```diff
@@ -660,8 +883,104 @@
...
+            </dd>
           </dl>
         </section>
+        <h3>
+          <dfn><code>DistortionCoefficients</code></dfn> dictionary
+        </h3>
+        <div data-dfn-for="DistortionCoefficients">
+          <dl class="dictionary-members">
+            <dt>
+              The <dfn>k1</dfn>, <dfn>k2</dfn>, <dfn>p1</dfn>, <dfn>p2</dfn>
```
anssiko, line 943: Without further context unclear what the semantics of the coefficients are.

index.src.html
```diff
@@ -660,8 +883,104 @@
...
+              <div class="note">
+                The first four elements of the array correspond to the first
+                matrix row, followed by four elements of the second matrix row
+                and so on. It is in format suitable for use with WebGL's
+                uniformMatrix4fv.
+              </div>
+            </dd>
+            <dt>
+              The <dfn><code>videoDeviceId</code></dfn> dictionary member
+              represents the deviceId of video camera the depth stream could be
```
anssiko: must be syncronized with

index.src.html
```diff
@@ -660,8 +883,104 @@
...
+              synchronized with. The deviceId can be used as a constraint in
```
anssiko: Make "The deviceId ..." a note.

index.src.html
```diff
@@ -778,6 +1097,15 @@
             requirements, since it might not be possible to synchronize tracks
             from sources.
           </p>
+          <p class="note">
+            This approach is simple to use but comes with the following
+            caveats: it might not be available and the resolutions of two
```
anssiko: might not be supported by the implementation

index.src.html
```diff
@@ -778,6 +1097,15 @@
...
+            tracks are intended to be the same. The alternative approach is
```
anssiko: same that can require downscaling and degrade quality.

index.src.html
```diff
@@ -778,6 +1097,15 @@
...
+            that a web developer implements, following the <a>algorithm to map
+            depth pixels to color pixels</a> and the corresponding <a href=
+            "https://www.w3.org/wiki/Media_Capture_Depth_Stream_Extension#Examples">
```
anssiko: Link to the example section, not to the wiki.

Suggested rewording: 

The alternative approach is that a web developer implements the algorithm to map depth pixels to color pixels in WebGL vertex shader.

index.src.html
```diff
@@ -842,16 +1170,71 @@
...
         </table>
         <p>
-          The <code>depthNear</code> and <code>depthFar</code> constrainable
-          properties, when set, allow the implementation to pick the best depth
-          camera mode optimized for the range <code>[depthNear,
-          depthFar]</code> and help minimize the error introduced by the lossy
-          conversion from the depth value <var>d</var> to a quantized
-          d<sub>8bit</sub> and back to an approximation of the depth value
-          <var>d</var>.
+          The <code>near</code> and <code>far</code> constrainable properties,
```
anssiko: depthNear and depthFar

index.src.html
astojilj: Note added:

> Radial distortion coefficients and tangential distortion coefficients are used when there's need to deproject depth value to 3D space or to project 3D value to 2D video frame coordinates.
> See algorithm to map depth pixels to color pixels and Brown-Conrady distortion model implementation 3D point cloud rendering example GLSL shader.
> 
> 

Don't have a good reference to Brown-Conrady model . Wikipedia [link](https://en.wikipedia.org/wiki/Distortion_(optics)#Software_correction)? 

Also, I feel we will need to add examples for the cases listed by @bbernhar in #110 but that would need to come later.

index.src.html
astojilj: It is now like you suggested:
I have omitted these two reasons - they could be even more important that the quality but maybe they shouldn't be here:
1. bandwidth and performance issue
The API is not good because it, by the design to be used on canvas2D, it implies significant performance hit: 
- to get downsampled content for use on canvas, we'll need to get the texture back from a) GPU to CPU (if sync and dowscalling happens there) or do the b) sync of two on CPU.
either way, sub optimal implementation.

2. As @bbernhar argumented in #110, it is also about the lost of precision  required for recognition use cases.

IMHO, we need to try different API approach for canvas2D.

index.src.html
astojilj: Thanks, removed the part "in WebGL vertex shader." as for recognition use cases, CPU implementation could be valid approach too (e.g. when processing 1080p color frame developer is interested in depth value for the specific pixel and doesn't necessary require full color frame processing).

astojilj:
PR Updated
@anssiko PTAL.

Open issue:

Added this note:

> Radial distortion coefficients and tangential distortion coefficients are used when there's need to deproject depth value to 3D space or to project 3D value to 2D video frame coordinates.
> See algorithm to map depth pixels to color pixels and Brown-Conrady distortion model implementation 3D point cloud rendering example GLSL shader.
> 
> 

Don't have a good reference to Brown-Conrady model . Wikipedia [link](https://en.wikipedia.org/wiki/Distortion_(optics)#Software_correction)?

anssiko:
Thanks @astojilj! I'll land this now and do some editorial cleanup afterwards.

PR #146, opened 2016-11-25
Fix #135: Replace fovs by focalLengthX and focalLengthY.

### LINKED ISSUE #135: Focal length seems under defined

astojilj:
@huningxin PTAL.
Thanks.

### DIFF
--- a/index.html
+++ b/index.html
@@ -420,19 +420,14 @@
           double. It represents the maximum range in meters.
         </p>
         <p>
-          A <a>depth map</a> has an associated <dfn>focal length</dfn> which is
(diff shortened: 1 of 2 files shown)


huningxin:
Thanks for this PR! LGTM. @robman @fluffy, Please take a look.

huningxin:
As it aligns the consensus made in #135. I'd like to merge this PR. For any follow ups, please open new issues to track. Thanks @astojilj !

PR #148, opened 2016-12-29
Replace near/far by depthNear/depthFar. Fix compilation on Windows.

astojilj:
Reason:
Generated C++ code from IDL cannot compile on Windows since near
and far are reserved keywords MS compiler.

It might be possible to undef it in C++ headers but this seems more
appropriate since:

MediaTrackSettings and other dictionaries are generic, meaning that
they apply to all stream tracks and it is good to explicitelly mark
that this applies only to depth. On the other side focalLength,
principal point and other intrinsics that we plan to add apply to
all cameras so there is no need to prefix them with depth.

This way, it is also in sync with depthNear and depthFar in WebVR
API's depthNear/depthFar[1].

[1]
https://w3c.github.io/webvr/#dom-vrdisplay-depthnear

### DIFF
--- a/index.html
+++ b/index.html
@@ -909,20 +909,20 @@
         </p>
         <pre class="idl">
           partial dictionary MediaTrackSettings {
-              double              near;
-              double              far;
+              double              depthNear;
+              double              depthFar;
               double              focalLengthX;
               double              focalLengthY;
           };
 </pre>
         <div dfn-for="MediaTrackSettings">
           <p>
-            The <dfn><code>near</code></dfn> dictionary member represents the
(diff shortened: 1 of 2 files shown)


astojilj:
@huningxin , PTAL.
I started implementing this in Chromium but generated code from IDL (V8, Blink modules code) couldn't compile on Windows. This fixes it.

For the reference, Chromium patch is ready for review https://codereview.chromium.org/2606983002/ and some feedback here would be needed in in order to progress with the patch review. Thanks.

huningxin:
LGTM. Thanks for the PR!

PR #149, opened 2017-01-23
Change depth to MediaTrackConstraints

### LINKED ISSUE #143: Change depth to MediaTrackConstraints
Current spec defines the depth as a new media type in `MediaStreamConstraints`

```
partial dictionary MediaStreamConstraints {
    (boolean or MediaTrackConstraints) depth = false;
};
```

However actually the depth camera is a kind of video capture device, it shares most of characteristics of a color camera. The major difference with a color camera is the pixel format and the meta data describing the pixel value (depth value).

So I propose to change the depth to the `MediaTrackConstraints`. To accommodate more kinds of video capture device, it would introduce the `videoKind` constrainable property:

```
partial dictionary MediaTrackConstraintSet {
             ConstrainDOMString videoKind;
};

enum VideoDeviceKindEnum {
    "color",
    "depth",
    "infrared",
    "fisheye"
};
```

The example of requesting a single depth track:
```
navigator.mediaDevices.getUserMedia({
  video: {videoKind: {exact: "depth"}}
}).then(function (stream) {
  // Render the depth video.
  var video = document.querySelector('#depth-video');
  video.srcObject = stream;
  video.play()
});
```

The example of requesting the color and depth tracks from one 3D camera by setting the `groupId` of the 3D camera.
```
navigator.mediaDevices.getUserMedia({
  video: {groupId: {exact: id}}
}).then(function (stream) {
  // Render the color video.
  var video = document.querySelector('#color-video');
  video.srcObject = stream;
  video.play()
});

navigator.mediaDevices.getUserMedia({
  video: {videoKind: {exact: "depth"}, groupId: {exact: id}}
}).then(function (stream) {
  // Render the depth video.
  var video = document.querySelector('#depth-video');
  video.srcObject = stream;
  video.play()
});
```

@robman @anssiko @fluffy @astojilj,  WDYT?

huningxin:
Fix #143

### DIFF
--- a/index.html
+++ b/index.html
@@ -1278,5 +1278,6 @@ gl.texImage2D(
         Project Tango for their experiments.
       </p>
     </section>
+    <section id="idl-index" class="appendix"></section>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -165,7 +165,7 @@
         This specification <a href=
         "https://w3c.github.io/mediacapture-main/#extensibility">extends</a>
         the <em>Media Capture and Streams</em> specification [[!GETUSERMEDIA]]
-        to allow a <a>depth-only stream</a> or combined <a>depth+video
+        to allow a <a>depth-only stream</a> or combined <a>depth+color
         stream</a> to be requested from the web platform using APIs familiar to
         web authors.
       </p>
@@ -210,7 +210,7 @@
         objects in the scene for that point in the array. A
         <code><a>MediaStream</a></code> object that contains one or more
         <a>depth stream track</a>s is referred to as a <a>depth-only stream</a>
-        or <a>depth+video stream</a>.
+        or <a>depth+color stream</a>.
       </p>
       <p>
         Depth cameras usually produce 16-bit depth values per pixel, so this
@@ -260,8 +260,6 @@
         The <code><a href=
         "https://w3c.github.io/mediacapture-main/archives/20160513/getusermedia.html#idl-def-Constraints">
         <dfn>Constraints</dfn></a></code>, <code><a href=
-        "https://w3c.github.io/mediacapture-main/archives/20160513/getusermedia.html#idl-def-MediaStreamConstraints">
-        <dfn>MediaStreamConstraints</dfn></a></code>, <code><a href=
         "https://w3c.github.io/mediacapture-main/archives/20160513/getusermedia.html#idl-def-MediaTrackSettings">
         <dfn>MediaTrackSettings</dfn></a></code>, <code><a href=
         "https://w3c.github.io/mediacapture-main/archives/20160513/getusermedia.html#idl-def-MediaTrackConstraints">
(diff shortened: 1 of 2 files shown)


huningxin:
@anssiko @robman @astojilj, PTAL.

BTW, I don't quite get the respec errors in the travis-ci:
```
😱 ReSpec error: Multiple <dfn>s for videoKind in MediaTrackConstraintSet
```

I actually only define `videoKind` once in `MediaTrackConstraintSet`. Did I miss anything?

anssiko:
Thanks for the PR!

It seems the remaining CI errors are false negatives due to a ReSpec limitation: ReSpec complains if the same dictionary member names are used and defined across multiple dictionaries. If we'd leave the `<dfn>`s out from other dictionaries, I believe ReSpec would give an error "`No <dfn> for XXXXX`".

anssiko:
I just validated my hypothesis: if we drop the other `<dfn>`s for `videoKind` we'll get the following warnings:
```
No <dfn> for videoKind in MediaTrackCapabilities.
No <dfn> for videoKind in MediaTrackConstraintSet.
No <dfn> for videoKind in MediaTrackSettings.
```

The same applies to the following members too defined in multiple dictionaries (the CI check halts on the first error so these go unnoticed):
```
depthNear
depthFar
focalLengthX
focalLengthY
```

Given "`No <dfn> for XXXXX`" warnings is better than the current errors, I suggest we keep `<dfn>`s for the above dictionary members only in `MediaTrackSupportedConstraints` and remove `<dfn>`s from any other dictionaries as we figure out how to fix this properly.

That said, the warnings would not block this PR from being merged.

anssiko:
@huningxin I figured this out. The proper fix for this is to use the `data-dfn-for` attribute as follows: `<dfn data-dfn-for="MediaTrackCapabilities"><code>videoKind</code></dfn>` and the same for the other dictionaries and members.

See the ReSpec docs: https://github.com/w3c/respec/wiki/data-dfn-for

Could you update this PR accordingly?

index.html
```diff
@@ -683,115 +681,396 @@
       </h2>
       <section>
         <h2>
-          <code><a>MediaStreamConstraints</a></code> dictionary
```
anssiko, line 686: We no longer extend `MediaStreamConstraints` so we can drop that from the Dependencies section.

anssiko:
@huningxin When you've done the updates, could you push the updated `index.src.html` too?

To build `index.html` from `index.src.html`, just `$ node build.js`. (We should probably document this workflow somewhere :-) - edit: documented the workflow at https://github.com/w3c/mediacapture-depth#pull-requests)

index.html
huningxin: Fixed.

huningxin:
> @huningxin I figured this out. The proper fix for this is to use the data-dfn-for attribute as follows: <dfn data-dfn-for="MediaTrackCapabilities"><code>videoKind</code></dfn> and the same for the other dictionaries and members.

I tried `data-dfn-for`. It doesn't solve the errors. I actually think it is an respec bug, as I already define the `data-dfn-for` in the 
```
<dl data-link-for="MediaTrackSupportedConstraints" data-dfn-for=
          "MediaTrackSupportedConstraints" class="dictionary-members">
```

It follows the mediacapture-main: https://github.com/w3c/mediacapture-main/blob/master/getusermedia.html#L1336

huningxin:
> @huningxin When you've done the updates, could you push the updated index.src.html too?
> 
> To build index.html from index.src.html, just $ node build.js. (We should probably document this workflow somewhere :-) - edit: documented the workflow at https://github.com/w3c/mediacapture-depth#pull-requests)

@anssiko , thanks for the instructions. I updated the index.src.html and build to index.html.

Please take another look.

huningxin:
The warnings:
```
line 503 column 89 - Warning: trimming empty <span>
line 504 column 42 - Warning: trimming empty <span>
line 605 column 92 - Warning: trimming empty <span>
line 606 column 48 - Warning: trimming empty <span>
```

It is introduced by MathJax generation code.

anssiko:
@huningxin Here's a PR to fix all the ReSpec errors: https://github.com/huningxin/mediacapture-depth/pull/1

LGTM otherwise. I'll merge this PR when you land the above PR, unless someone else has comments.

(We don't need to worry about the MathJax warnings.)

huningxin:
> @huningxin Here's a PR to fix all the ReSpec errors: huningxin#1

@anssiko , thanks much for the fix! It is merged.

anssiko:
Merged, and the [Editor's Draft][1] is updated accordingly.

@huningxin Thanks for fixing this major issue!

[1]: https://w3c.github.io/mediacapture-depth/

PR #150, opened 2017-01-23
Add contribution guidelines to README

anssiko:


anssiko:
Noticed we hadn't documented the workflow that's a bit unusual given it includes a custom build step. That's now fixed.

PR #151, opened 2017-01-24
Fix warnings of node build.js

huningxin:


huningxin:
Will combine this to https://github.com/w3c/mediacapture-depth/pull/149. Close it.

PR #152, opened 2017-01-31
Use data-cite for automatic cross-referencing

anssiko:
Use data-cite ReSpec feature for automatic cross-referencing:
https://lists.w3.org/Archives/Public/spec-prod/2017JanMar/0011.html

This feature was very much welcome for this spec that had a long list of dependencies. Thanks @marcoscaceres!

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,108 +245,90 @@
         Dependencies
       </h2>
       <p>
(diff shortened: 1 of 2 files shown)


anssiko:
Editorial, merging without review.

Issue #57, opened 2015-01-23
Improve examples


kenchris:
According to Marcos it is fine to use ES15 in specs, like 'let' in loops and => instead of functions.

Also s/reason/err/ in the promise catch. And write out gUM. it is not obvious what that is, and examples should be obvious

anssiko:
@astojilj promised to improve the examples inline with the API changes that will land soon.

anssiko:
Could reuse some examples for the @astojilj's article: https://01.org/chromium/blogs/astojilj/2017/depth-camera-capture-html5

I suggest checking any example code we add to the spec with http://jshint.com/.

kenchris:
The second example is impossible to follow as it only does half of what it says and just contains a lot of ... :-)

https://github.com/01org/depth-camera-web-demo/blob/master/depthdemo.html does more or less what it says, but contains optimizations using extensions etc. If that could be simplified somehow we might be able to include it some how...

Shaders could be multiline using ``

```
gl.shaderSource(pixel_shader, `
  precision mediump float;

  uniform sampler2D s;
  varying vec2 t;

  void main() {
    vec4 tex = texture2D(s, t); 
    gl_FragColor = vec4(tex.r, tex.g, tex.b, tex.a);
  }`);

```

astojilj:
@kenchris 
Unfortunately, seems that the second example is not going to be supported (upload 16 bit data to split R&G components).

Usage of floats mandates activating via extension (otherwise no way to use float and it is the only and seems like the best way):
in WebGL2: gl.getExtension('EXT_color_buffer_float');
in WebGL1: gl.getExtension('OES_texture_float'); (I also need to update the demo - thanks for pointing this out)

In WebGL2, EXT_color_buffer_float also brings a possibility to use single component float, with 4x lower bandwidth and much faster upload to texture consumption.

I don't know about the platform where these extensions are not supported.

Good point about multiline, will upload the demo. Thanks.

kenchris:
@astojilj I guess we could change the example then to more or less what you are doing in your demo.

I think for the sake of the spec, showing webgl2 should be fine. It is an example after all and it will be widely supported (Safari said they will support it too in addition to their WebGPU work)

PR #154, opened 2017-02-22
Improve examples a bit

kenchris:
Move to ES2017 and make the examples a bit easier to understand

***
[Preview](https://s3.amazonaws.com/pr-preview/kenchris/mediacapture-depth/gh-pages.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/mediacapture-depth/573f01e...kenchris:a9f294e.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1291,79 +1291,109 @@
       <h2>
         Examples
       </h2>
-      <h3>
-        Playback of depth and color streams from same device group.
-      </h3>
-      <pre class="example">navigator.mediaDevices.getUserMedia({
-  video: {videoKind: {exact: "color"}, groupId: {exact: id}}
-}).then(function (stream) {
-    // Wire the media stream into a &lt;video&gt; element for playback.
-    // The RGB video is rendered.
-    var video = document.querySelector('#video');
-    video.srcObject = stream;
-    video.play();
+      <aside title="Playback of depth and color streams from same device group" class="example">
+        <pre class="highlight">async function attachVideoStream(el, kind, groupId) {
+  const constraints = {
+    video: {
+      mandatory: {
+        groupId: groupId
+      },
+      videoKind: { exact: kind }
+    }
   }
-);
 
-navigator.mediaDevices.getUserMedia({
-  video: {videoKind: {exact: "depth"}, groupId: {exact: id}}
-}).then(function (stream) {
-    // Wire the depth-only stream into another &lt;video&gt; element for playback.
-    // The depth information is rendered in its grayscale representation.
-    var depthVideo = document.querySelector('#depthVideo');
-    depthVideo.srcObject = stream;
-    depthVideo.play();
-  }
-);
+  const stream = await navigator.mediaDevices.getUserMedia(constraints);
+
+  el.srcObject = stream;
+  el.play();
+
+  return el;
+}
+
+async function play() {
+  const colorEl = document.createElement("video");
+  const depthEl = document.createElement("video");
+
+  const body = document.querySelector('body');
+  body.appendChild(colorEl);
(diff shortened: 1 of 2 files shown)


anssiko:
Marked as non-substantive for IPR from ash-nazg.

anssiko:
Thanks @kenchris! (Fixes #57.)

PTAL @astojilj @huningxin.

anssiko:
(No need to worry about the build failure, it is a harmless warning of empty `<span>`s caused by MathJax integration used for formulas.)

index.html
```diff
@@ -1213,59 +1216,94 @@
...
+
+async function play() {
+  const colorEl = document.createElement("video");
+  const depthEl = document.createElement("video");
+
+  const body = document.querySelector('body');
+  body.appendChild(colorEl);
+  body.appendChild(depthEl);
+
+  // Assume that all our video inputs are depth stream capable.
```
astojilj, line 1321: We can use videoKind to avoid the need for this assumptions - usually there is an integrated camera and this approach (via navigator.mediaDevices.enumerateDevices()) didn't work. I guess counting the MediaDeviceInfo that are videoinput with the same groupId would give a closer match but it still looks like a hack.
  
We shouldn't need to use enumerateDevices - first use videoKind to get depth stream,  then get groupId from the depthStream's track and then use videoKind: color + groupId to get corresponding color device.
I didn't try this yet - it depends on experimental MediaStreamTrack.getSettings() - https://crbug.com/681824. Maybe it is already available through MediaTrackCapabilities though.

The implementation for videoKind has recently been added to Chromium: https://crrev.com/2664673002/.

index.html
```diff
@@ -1213,59 +1216,94 @@
...
-  video: {videoKind: {exact: "depth"}, groupId: {exact: id}}
-}).then(function (stream) {
-    // Wire the depth-only stream into another &lt;video&gt; element for playback.
-    // The depth information is rendered in its grayscale representation.
-    var depthVideo = document.querySelector('#depthVideo');
-    depthVideo.srcObject = stream;
-    depthVideo.play();
-  }
-);
+  const stream = await navigator.mediaDevices.getUserMedia(constraints);
```
astojilj, line 1305: I like await. Is it a problem that it looks different compared to original approach in main spec?
https://w3c.github.io/mediacapture-main/#examples

index.html
```diff
@@ -1213,59 +1216,94 @@
...
-  video: {videoKind: {exact: "depth"}}
-}).then(function (stream) {
-  // wire the stream into a &lt;video&gt; element for playback
-  var depthVideo = document.querySelector('#depthVideo');
-  depthVideo.srcObject = stream;
-  depthVideo.play();
-}).catch(function (reason) {
-  // handle gUM error here
-});
+      </aside>
```
astojilj, line 1338: This approach is going to be deleted soon - reconstruction from split components is not supported.
We need to base the example on floats - like in https://software.intel.com/en-us/blogs/2017/02/10/depth-camera-capture-in-html5.

astojilj (commented):
@kenchris Thanks for help on this.

index.html
kenchris: I would say no. This spec is newer and thus will use newer platform features.

index.html
kenchris: Can you try to modify the example and test it there (I dont have the hardware), then I will update the example

index.html
kenchris: Let's change it to your demo as example instead then.

kenchris: I can try to help simplify that as well but I cannot test here

kenchris:
Rebased, will have to look at this later... comments are of course always welcome

PR #155, opened 2017-03-03
WebGL section initial content. Updated the examples.

astojilj:
Renoves the splitting depth to red and green component.
@anssiko @huningxin PTAL


anssiko:
`depthTexture` is undefined, can we define its creation for completeness?

astojilj:
> depthTexture is undefined, can we define its creation for completeness?
Fixed in updated PR.

anssiko:
@kenchris please review.

index.html
```diff
@@ -1194,16 +1197,102 @@
             This section is currently work in progress, and subject to change.
           </div>
           <p>
+            Use cases like rendering 3D point cloud, background removal,
```
kenchris: I would do

There are several use-cases which are a good fit to be, at least partially, implemented on the GPU, such as motion recognition, pattern recognition, background removal, as well as 3D point cloud.

This section explains which APIs can be used for some of these mentioned use-cases; ...

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+          <h3>
+            Upload video frame to WebGL texture
+          </h3>
+          <p>
             A <a>video</a> element whose source is a <a>MediaStream</a> object
             containing a <a>depth stream track</a> may be uploaded to a WebGL
-            texture of format <code>RGB</code> and type
-            <code>UNSIGNED_BYTE</code>. [[WEBGL]]
+            texture of format <code>RGBA</code> or <code>RED</code> and type
+            <code>FLOAT</code>. See also the specification [[WEBGL]] and the
```
kenchris: See also sounds a bit weird to me. Maybe just remove also?

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            "padding-top: 0.372em; padding-bottom: 0.298em;">t</span></span></span></span></span></span><span class="mjx-denominator"
+            style="width: 3.478em; bottom: -0.778em;"><span id="MJXc-Node-118"
+            class="mjx-mn"><span class="mjx-char MJXc-TeX-main-R" style=
+            "padding-top: 0.372em; padding-bottom: 0.372em;">65535.0</span></span></span><span style="border-bottom: 1.3px solid; top: -0.296em; width: 3.478em;"
+            class="mjx-line"></span></span><span style=
+            "height: 2.225em; vertical-align: -0.778em;" class=
+            "mjx-vsize"></span></span></span></span></span></span></span>
+          </p>
+          <h3>
+            Read the data from WebGL texture
```
kenchris: from *a*

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+          </h3>
+          <p>
+            We list here some of the possible approaches:In addition to , there
```
kenchris: Here we list some of the possible approaches. (s/:/. and remove space before ,)

kenchris: in addition to what?

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            are two ways available for asynchronous access:
+          </p>
+          <ul>
+            <li>Synchronous readPixels requires the least amount of code and it
```
kenchris: You said async above and now talk about sync... 

Do you mean: In addition to asynchronous access, there are ....

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            is available om WebGL1. See the <a>readPixels from float</a>
+            example for details.
```
kenchris: for further details?

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            </li>
+            <li>Asynchronous readPixels using pixel buffer objects should
+            remove the blocking while waiting for the readPixels call.
```
kenchris: the -> any blocking?

waiting for the readPixels call to finish?

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            </li>
+            <li>Transform feedback with GetBufferSubData(Async) provides
+            synchronos and asynchronous read access to depth and color texture
```
kenchris: spellign of synchronous

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            data processed in vertex shader.
```
kenchris: in *the* vertex shader

index.html
```diff
@@ -1194,16 +1197,102 @@
...
+            </li>
+          </ul>
+          <p></p>
+          <p class="note">
+            Performance of synchronous <a>readPixels from float</a> example in
```
kenchris, line 1045: a bit hard to read.

The "readPixel from float" example above, while being simplistic, suffers from some performance issues, due to the fact that ...  This can be mitigated by...

index.html
```diff
@@ -1239,12 +1328,11 @@
...
       <h3>
-        WebGL Fragment Shader based post-processing
+        WebGL2: <dfn>upload to float texture</dfn>
       </h3>
       <pre class="example">
-// This code sets up a video element from a depth stream, uploads it to a WebGL
-// texture, and samples that texture in the fragment shader, reconstructing the
-// 16-bit depth values from the red and green channels.
+// This code sets up a video element from a depth stream, uploads it to a WebGL2
+// float texture holding the full precision data as the 16-bit depth map.
```
kenchris: holding -> containing?

index.html
```diff
@@ -1256,31 +1344,78 @@
   // handle gUM error here
 });
 
+// ... initialize WebGL2 context when application starts.
```
kenchris: pretty useless comment, if people dont know that, they cannot understand any example here :) it is basic knowledge of webgl. Comments should make sense, or dont have them

index.html
```diff
@@ -1256,31 +1344,78 @@
...
-&lt;/script&gt;
 </pre>
+      <h3>
+        WebGL2: <dfn>readPixels from float</dfn> texture
+      </h3>
+      <p>
+        This example extends <a>upload to float texture</a> example.
+      </p>
+      <pre class="example">
+// This code sets up a named framebuffer, attach the texture we upload the depth
```
kenchris: attach the texture we upload the depth - sounds weird

index.html
```diff
@@ -1256,31 +1344,78 @@
...
+// video to as color attachment and, in rendering time, reads the texture
```
kenchris: AT rendering time

index.html
```diff
@@ -1256,31 +1344,78 @@
...
+// content  to Float32Array.
+
+// ... initialize texture and framebuffer for reading back the texture...
```
kenchris: why the ... at the beginning, what purpose do they serve?

index.html
```diff
@@ -1256,31 +1344,78 @@
...
+gl.bindTexture(gl.TEXTURE_2D, depthTexture);
+gl.texImage2D(
+   gl.TEXTURE_2D,
+   0,
+   gl.R32F,
+   gl.RED,
+   gl.FLOAT,
+   depthVideo);
+
+var buffer = new Float32Array(depthVideo.videoWidth * depthVideo.videoHeight);
```
kenchris: const should work instead of var here... var is not recommended in JS now. const doesnt make the object const, only the reference which cannot be reassigned.

index.html
astojilj: thanks

index.html
astojilj: thanks

index.html
astojilj: thanks

index.html
astojilj: thanks. should have remove it. Now the content is only: 
Here we list some of the possible approaches.
<list entries>

index.html
astojilj: removed the mention of async above.

index.html
astojilj: thanks.

index.html
astojilj: Replaced by: Asynchronous readPixels using pixel buffer objects to avoid blocking the readPixels call.

index.html
astojilj: thanks

index.html
astojilj: thanks.

index.html
astojilj: I wanted to say: blocking is not bad, it is only 3-6 ms on my development laptops.

index.html
astojilj: thx

index.html
astojilj: ok, removed

index.html
astojilj: Changed my mind about: "the texture we upload the depth video to" -> "the texture with depth data". Doesn't have the depth data when created.
 
Now:
// This code creates the texture to which we will upload the depth video frame.
// Then, it sets up a named framebuffer, attach the texture as color attachment

index.html
astojilj: thanks. all replaced by:
// This code creates the texture to which we will upload the depth video frame.
// Then, it sets up a named framebuffer, attach the texture as color attachment
// and, after uploading the depth video to the texture, reads the texture
// content to Float32Array.

index.html
astojilj: Don't have a clue. It was there before.
I am often comfortable accepting existing code approach and replicating it without understanding what is it for.

index.html
astojilj: OK. It is better to pre-allocate the buffer outside rendering loop, and we can do it only after texImage2D and gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_FORMAT) call.
That's why it is var declared somewhere else but initialized here.
It complicates the example so I will use const and put the comment.

astojilj:
@kenchris 
Thanks. Updated. PTAL.

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+            There are several use-cases which are a good fit to be, at least
+            partially, implemented on the GPU, such as motion recognition,
+            pattern recognition, background removal, as well as 3D point cloud.
+          </p>
+          <p>
+            This section explains which APIs can be used for some of these
+            mentioned use-cases; the concrete usage examples are provided in
+            the <a href=
+            "https://www.w3.org/wiki/Media_Capture_Depth_Stream_Extension#Examples">
+            Examples</a> section.
```
kenchris: section below?

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+          </p>
+          <h3>
+            Upload video frame to WebGL texture
```
kenchris: a video frame, or video frames?

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+          </h3>
+          <p>
             A <a>video</a> element whose source is a <a>MediaStream</a> object
             containing a <a>depth stream track</a> may be uploaded to a WebGL
-            texture of format <code>RGB</code> and type
-            <code>UNSIGNED_BYTE</code>. [[WEBGL]]
+            texture of format <code>RGBA</code> or <code>RED</code> and type
```
kenchris, line 1015: Not BLUE? If not, explain why. if yes, then change the wording

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+            "padding-top: 0.372em; padding-bottom: 0.298em;">t</span></span></span></span></span></span><span class="mjx-denominator"
+            style="width: 3.478em; bottom: -0.778em;"><span id="MJXc-Node-118"
+            class="mjx-mn"><span class="mjx-char MJXc-TeX-main-R" style=
+            "padding-top: 0.372em; padding-bottom: 0.372em;">65535.0</span></span></span><span style="border-bottom: 1.3px solid; top: -0.296em; width: 3.478em;"
+            class="mjx-line"></span></span><span style=
+            "height: 2.225em; vertical-align: -0.778em;" class=
+            "mjx-vsize"></span></span></span></span></span></span></span>
+          </p>
+          <h3>
+            Read the data from a WebGL texture
```
kenchris: I am wondering whether people say "of a texture" - so probably check with a native speaker

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+          </h3>
+          <p>
+            Here we list some of the possible approaches.
+          </p>
+          <ul>
+            <li>Synchronous readPixels requires the least amount of code and it
```
kenchris: Synchronous calls to <code>readPixels</code>, which requires...

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+            is available om WebGL1. See the <a>readPixels from float</a>
```
kenchris: om? works with WebGL 1.0 (that is the spec name :))

readPixels should be with <code> right?

index.html
```diff
@@ -1194,16 +1197,104 @@
...
+            example for further details.
+            </li>
+            <li>Asynchronous readPixels using pixel buffer objects to avoid
+            blocking the readPixels call.
+            </li>
+            <li>Transform feedback with GetBufferSubData(Async) provides
```
kenchris: I am not sure it is clear what "transform feedback" is, or whether it is one concept... can we add link or at least make it cursive or so

index.html
kenchris: But web vr and such use cases will be on entirely different hardware, so not sure that we can make such a statement

index.html
```diff
@@ -1239,12 +1330,11 @@
 );
 </pre>
       <h3>
-        WebGL Fragment Shader based post-processing
+        WebGL2: <dfn>upload to float texture</dfn>
```
kenchris: Spec is called WebGL 2.0

index.html
```diff
@@ -1239,12 +1330,11 @@
...
       </h3>
       <pre class="example">
-// This code sets up a video element from a depth stream, uploads it to a WebGL
-// texture, and samples that texture in the fragment shader, reconstructing the
-// 16-bit depth values from the red and green channels.
+// This code sets up a video element from a depth stream, uploads it to a WebGL2
+// float texture containing the full precision data as the 16-bit depth map.
```
kenchris: How is this depth map exposed, Uint16Array?

index.html
```diff
@@ -1256,31 +1346,80 @@
   // handle gUM error here
 });
 
-// ... later, in the rendering loop ...
+var gl = canvas.getContext("webgl2");
+// Activate the extension to use single component R32F texture format.
```
kenchris: Is that alwasy available in WebGL 2.0? if so I would write 

Activate the standard WebGL 2.0 extension for using single component R32F texture format

index.html
```diff
@@ -1256,31 +1346,80 @@
...
+gl.getExtension('EXT_color_buffer_float');
+
+// Later, in the rendering loop ...
```
kenchris: Could be have this be in another box in the spec, so it is super clear that this is a separate part of the code

index.html
```diff
@@ -1256,31 +1346,80 @@
...
-    vec4 floatColor = texture2D(u_tex, v_texCoord);
-    float dn = floatColor.r;
-    float depth = 0.;
-    depth = far * near / ( far - dn * ( far - near));
-    // ...
-  }
-&lt;/script&gt;
 </pre>
+      <h3>
+        WebGL2: <dfn>readPixels from float</dfn> texture
```
kenchris: <code>

index.html
```diff
@@ -1256,31 +1346,80 @@
...
+      </h3>
+      <p>
+        This example extends <a>upload to float texture</a> example.
+      </p>
+      <pre class="example">
+// This code creates the texture to which we will upload the depth video frame.
```
kenchris: This should not be in the code comments... check my former patch to see how it can be done as part of the example box

index.html
```diff
@@ -1256,31 +1346,80 @@
...
+// Then, it sets up a named framebuffer, attach the texture as color attachment
+// and, after uploading the depth video to the texture, reads the texture
+// content to Float32Array.
+
+// Initialize texture and framebuffer for reading back the texture...
```
kenchris: Sometimes it is better to define functions with descriptive names instead of comments. It also makes the code nicely scoped and easy to read

index.html
```diff
@@ -1256,31 +1346,80 @@
...
+var depthTexture = gl.createTexture();
```
kenchris: const

index.html
```diff
@@ -1256,31 +1346,80 @@
...
+gl.bindTexture(gl.TEXTURE_2D, depthTexture);
+gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
+
+var framebuffer = gl.createFramebuffer();
```
kenchris: const

index.html
```diff
@@ -1256,31 +1346,80 @@
...
+gl.bindTexture(gl.TEXTURE_2D, depthTexture);
+gl.texImage2D(
+   gl.TEXTURE_2D,
+   0,
+   gl.R32F,
+   gl.RED,
+   gl.FLOAT,
+   depthVideo);
+
+// It is recommended to allocate this buffer once.
```
kenchris: only once? I dont think the comment is clear

index.html
astojilj: copying approach from other places, prefer not to have bellow together with link.

index.html
astojilj: There is no BLUE texture in webgl.

index.html
astojilj: video frames, since there is a mention of loop. thanks.

index.html
astojilj: from a texture memory, from a texture or read back the texture data is used. keeping the existing.

index.html
astojilj: replaced by:
if (!buffer) {
  buffer =
      new Float32Array(depthVideo.videoWidth * depthVideo.videoHeight);
}

index.html
astojilj: let - it is supposed to be used in different scope from creation. feel free to refactor in your patch

index.html
astojilj: let

index.html
astojilj: Please feel free to handle it in your PR #154. My goal with this patch is to replace the original example code - since it is not supported and need to be removed ASAP. Cosmetic changes will need to be handled in follow up.

index.html
astojilj: moved out to <p></p>. You are adding <aside> in PR #154 so it is better to handle it there for all. Thanks

index.html
astojilj: Replaced WebGL2 by WebGL.

index.html
astojilj: It is better if handled in your PR #154.

index.html
astojilj: removed the comment.
depth map is 16 bit. It is not exposed.
The texture is float.

index.html
astojilj: Links to khronos specifications added for both Transform feedback with GetBufferSubDatasync
Don't have a clue how to add informative reference to the bottom of the page [WEBGL2], @anssiko how to do that? Thanks

index.html
astojilj: Synchronous readPixels usage. It is the same call as async, just different setup.

astojilj:
PR updated.
@kenchris Thanks. I have incorporated the text changes and some of the suggested example code refactoring.
I think it is more productive if you continue improving the examples in PR #154 - here I covered just the high priority work: replace unsupported example code.

Open: 
Links to khronos specifications added for both Transform feedback with GetBufferSubDatasync
Don't have a clue how to add informative reference to the bottom of the page [WEBGL2], @anssiko how to do that? Thanks

kenchris:
Can you rebase this so it can get merged?

astojilj:
Rebased. Thanks.

anssiko:
@astojilj You can add your Khronos informative references to `localBiblio`, see https://github.com/w3c/mediacapture-depth/blob/gh-pages/index.src.html#L80

To make a reference informative in the prose, use `[[WEBGL2]]` i.e. without "!". `[[!WEBGL2]]` would make it normative.

astojilj:
@anssiko 
Added [WEBGL2] and [WEBGL-GET-BUFFER-SUB-DATA-ASYNC]. 
This PR is updated and no open issues to merge it.

I will start adding  [WEBGL2] and [WEBGL-GET-BUFFER-SUB-DATA-ASYNC] to [SPECREF](https://github.com/tobie/specref/blob/05d19e0c1a6d6cebc4ad2221b549e9e518e9a97d/refs/biblio.json)
as the usage of [localBiblio is discouraged](https://github.com/w3c/respec/wiki/localBiblio) and submit another PR once ready.

anssiko:
Thanks!

PR #158, opened 2017-03-14
Split dictionary members

anssiko:
To be iterated further, PTAL @astojilj

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/mediacapture-depth/split-dictionary-members.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/mediacapture-depth/573f01e...4e2f19c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -619,7 +619,7 @@
               double              principalPointY;
               DistortionCoefficients deprojectionDistortionCoefficients;
               DistortionCoefficients projectionDistortionCoefficients;
-              Transform depthToVideoTransform;
+              Transformation      depthToVideoTransform;
           };
 
           dictionary DistortionCoefficients {
--- a/index.html
+++ b/index.html
@@ -551,20 +551,27 @@
       <h2>
         Extensions
       </h2>
+      <p>
+        If the implementation is unable to report the value represented by any
+        of the dictionary members, they are not <a>present</a> in the
+        dictionary.
+      </p>
       <section>
         <h2>
           <a>MediaTrackSupportedConstraints</a> dictionary
         </h2>
         <pre class="idl">          partial dictionary MediaTrackSupportedConstraints {
+              // Apply to both depth stream track and color stream track:
               boolean videoKind = true;
-              boolean depthNear = true;
-              boolean depthFar = true;
               boolean focalLengthX = true;
               boolean focalLengthY = true;
               boolean principalPointX = true;
               boolean principalPointY = true;
               boolean deprojectionDistortionCoefficients = false;
               boolean projectionDistortionCoefficients = false;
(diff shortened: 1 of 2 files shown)


alvestrand:
@miguelao some of the constraints being worked on here seem to overlap with those in https://github.com/w3c/mediacapture-image - can you review?
(apologies for somewhat off-topic comment)

anssiko:
Note this PR is very much WIP, so I would be cautious to review it too carefully. Based on a quick scan of https://w3c.github.io/mediacapture-image/ there are no overlapping constraints per naming at least. Ping also @astojilj to confirm the details.

We're trying to figure out how to best spec the extensions to constrains pattern while keeping the spec concise and clear. Inheriting the boilerplate directly from mediacapture-main was our initial attempt but led to a spec that was hard to follow (and edit).

For example, in mediacapture-main, the following "Dictionary Foo Members" sections duplicate information already in the IDL definition, and add "See Baz for details." indirections that eventually all link to the same "Constrainable Properties" table:

https://www.w3.org/TR/mediacapture-streams/#idl-def-MediaTrackSupportedConstraints
https://www.w3.org/TR/mediacapture-streams/#idl-def-MediaTrackCapabilities
https://www.w3.org/TR/mediacapture-streams/#idl-def-MediaTrackConstraintSet
https://www.w3.org/TR/mediacapture-streams/#idl-def-MediaTrackSettings

Our plan is to link directly from the dictionary IDL members to the "Constrainable properties" section where we define the expected behaviour for each dictionary member per kind (color or depth). Probably we should get rid of the table as well since it is bad for readability, making the columns too narrow for text. Also "Notes" column contains normative assertions which is confusing.

I like the format in https://w3c.github.io/mediacapture-image/#constrainable-properties better over https://www.w3.org/TR/mediacapture-streams/#constrainable-properties, so we probably do something similar for this spec.

Recently we've been busy with the [Blink][1] implementation and have experimented with some [demos][2] using our implementation. We'll follow up with spec updates baking in our implementation feedback (including the editorial ones discussed above) hopefully soon.

[1]: https://crbug.com/616098
[2]: https://github.com/01org/depthcamera-pointcloud-web-demo

yell0wd0g:
@alvestrand I see no overlap constraints - could you please elaborate?

alvestrand:
@miguelao focalLengthX, focalLengthY, principalPointX, principalPointY were the ones that caught my eye. mediastream-image has pointsOfInterest and zoom.

Not sure if there's a need for alignment, but I thought it was worth checking out.

astojilj:
LGTM

anssiko:
Thanks @astojilj. Will land this PR, opened #159 for further improvements.

PR #160, opened 2017-03-23
Improve constrainable properties sections

### LINKED ISSUE #159: Improve constrainable properties sections
* We now link directly from the dictionary IDL members to the "Constrainable properties" sections where the expected behaviour for each dictionary member per kind (color or depth) should be defined. Ensure each of the constraints in these sections is defined properly.

* Editorial: convert the "constrainable properties" tables to simple prose. The tables make the columns too narrow for text, also "Notes" column contains normative assertions which is confusing. Prefer the format in https://w3c.github.io/mediacapture-image/#constrainable-properties over https://www.w3.org/TR/mediacapture-streams/#constrainable-properties.

anssiko:
Fix #159. PTAL @astojilj

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/mediacapture-depth/issue-159-constrainable-properties.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/mediacapture-depth/7b04774...5fe6430.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -112,6 +112,10 @@
 
     </script>
     
+    <style>
+/* workaround to hide redundant dfns */
+    p.related { visibility: hidden; height: 0px; }
+    </style>
   <style id="MathJax_CHTML_styles">.mjx-chtml {display: inline-block; line-height: 0; text-indent: 0; text-align: left; text-transform: none; font-style: normal; font-weight: normal; font-size: 100%; font-size-adjust: none; letter-spacing: normal; word-wrap: normal; word-spacing: normal; white-space: nowrap; float: none; direction: ltr; max-width: none; max-height: none; min-width: 0; min-height: 0; border: 0; margin: 0; padding: 1px 0}
 .MJXc-display {display: block; text-align: center; margin: 1em 0; padding: 0}
 .mjx-chtml[tabindex]:focus, body :focus .mjx-chtml[tabindex] {display: inline-table}
@@ -317,7 +321,14 @@
         interfaces this specification extends are defined in [[!GETUSERMEDIA]].
       </p>
       <p>
-        The <dfn data-cite="!GETUSERMEDIA#idl-def-Constraints"><code>Constraints</code></dfn>,
+        The concepts <dfn data-cite="!GETUSERMEDIA#dfn-constraints"><code>Constraints</code></dfn>,
+        <dfn data-cite="!GETUSERMEDIA#dfn-capabilities"><code>Capabilities</code></dfn>,
+        <dfn data-cite="!GETUSERMEDIA#idl-def-ConstraintSet"><code>ConstraintSet</code></dfn>,
+        and <dfn data-cite="!GETUSERMEDIA#dfn-settings"><code>Settings</code></dfn>, and
+        <dfn data-cite="!GETUSERMEDIA#types-for-constrainable-properties">types
+        of constrainable properties</dfn> are defined in [[!GETUSERMEDIA]].
+      </p>
+      <p>
         <dfn data-cite="GETUSERMEDIA#idl-def-MediaTrackSettings"><code>MediaTrackSettings</code></dfn>,
         <dfn data-cite="!GETUSERMEDIA#idl-def-MediaTrackConstraints"><code>MediaTrackConstraints</code></dfn>,
         <dfn data-cite="!GETUSERMEDIA#idl-def-MediaTrackSupportedConstraints"><code>MediaTrackSupportedConstraints</code></dfn>,
@@ -560,6 +571,12 @@
         <h2>
           <a>MediaTrackSupportedConstraints</a> dictionary
         </h2>
+        <p>
+          <a>MediaTrackSupportedConstraints</a> dictionary represents the list
+          of <a>Constraints</a> recognized by a <a>user agent</a> for
+          controlling the <a>Capabilities</a> of a <a>MediaStreamTrack</a>
+          object.
+        </p>
         <pre class="idl">          partial dictionary MediaTrackSupportedConstraints {
               // Apply to both depth stream track and color stream track:
               boolean videoKind = true;
@@ -580,6 +597,10 @@
         <h2>
           <a>MediaTrackCapabilities</a> dictionary
         </h2>
+        <p>
+          <a>MediaTrackCapabilities</a> dictionary represents the
+          <a>Capabilities</a> of a <a>MediaStreamTrack</a> object.
+        </p>
         <pre class="idl">          partial dictionary MediaTrackCapabilities {
               // Apply to both depth stream track and color stream track:
               DOMString videoKind;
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -317,7 +321,14 @@
         interfaces this specification extends are defined in [[!GETUSERMEDIA]].
       </p>
       <p>
-        The <dfn data-cite="!GETUSERMEDIA#idl-def-Constraints"><code>Constraints</code></dfn>,
+        The concepts <dfn data-cite="!GETUSERMEDIA#dfn-constraints"><code>Constraints</code></dfn>,
+        <dfn data-cite="!GETUSERMEDIA#dfn-capabilities"><code>Capabilities</code></dfn>,
+        <dfn data-cite="!GETUSERMEDIA#idl-def-ConstraintSet"><code>ConstraintSet</code></dfn>,
+        and <dfn data-cite="!GETUSERMEDIA#dfn-settings"><code>Settings</code></dfn>, and
```
astojilj, line 327: , and Setting, and types => , Settings and types

index.html
```diff
@@ -376,7 +387,8 @@
         are defined in [[WEBIDL]].
       </p>
       <p>
-        The meaning of dictionary member being <dfn data-cite="!WEBIDL#dfn-present">present</dfn> or <dfn data-cite="!WEBIDL#dfn-present">not present</dfn> is defined in [[WEBIDL]].
+        The meaning of dictionary member being <dfn data-cite="!WEBIDL#dfn-present">present</dfn> or <dfn data-cite="!WEBIDL#dfn-present">not present</dfn>, and its <dfn data-cite="!WEBIDL#dfn-dictionary-member-default-value">default value</dfn> are
```
astojilj, line 397: "being present or not present, and its default value are"  => "being present or not present and having a default value are"

index.html
```diff
@@ -600,6 +629,16 @@
...
           <code>MediaTrackConstraintSet</code> dictionary
         </h2>
+        <p>
+          <a>ConstraintSet</a> dictionary specifies each member's set of
+          allowed values.
+        </p>
+        <p>
+          See the <a>types of constrainable properties</a> for the definitions
+          of <dfn><code>ConstrainDOMString</code></dfn>,
+          <dfn><code>ConstrainDouble</code></dfn>, and
```
astojilj: Nit: ", and" => " and"

index.html
```diff
@@ -560,26 +572,43 @@
         <h2>
           <a>MediaTrackSupportedConstraints</a> dictionary
         </h2>
+        <p>
+          <a>MediaTrackSupportedConstraints</a> dictionary represents the list
```
astojilj, line 583: This sentence is the same as in main spec, but it is a bit difficult to grasp just from the sentence itself.
Propose to just keep the reference to main spec and add here:
"Partial dictionary <a>MediaTrackSupportedConstraints</a> extends the original dictionary defined in [GETUSERMEDIA]. Boolean value true represents applicable MediaStreamTrack constraint.

index.html
```diff
@@ -560,26 +572,43 @@
...
+          object.
+        </p>
         <pre class="idl">          partial dictionary MediaTrackSupportedConstraints {
               // Apply to both depth stream track and color stream track:
               boolean videoKind = true;
-              boolean focalLengthX = true;
-              boolean focalLengthY = true;
-              boolean principalPointX = true;
-              boolean principalPointY = true;
+              boolean focalLengthX = false;
```
astojilj, line 600: This defines the default value, but still allows the user agent to override and return true.
Ideally, we could have only one field in Capabilities and Constraints (videoKind) and have all of the fields in Settings, but it is not possible if we stick to definition in main spec:

Capabilities:
Capabilities is a dictionary containing one or more key-value pairs, where each key MUST be a constrainable property, and each value MUST be a subset of the set of values allowed for that property.
Settings:
"There MUST be a single value for each key and the value MUST be a member of the set defined for that property by getCapabilities()."

So, as we cannot have a field in Settings but no fields with the same name in Constraint/Dependencies we can just set default to false.
We cannot change  _(double or DoubleRange)_ focalLengthX; to _double_ focalLengthX in MediaTrackCapabilities (what we discussed yesterday).
Anyway, just adding a note, I have no objection to the change here.

index.html
```diff
@@ -560,26 +572,43 @@
...
+          dictionary is false, are omitted by the <a>user agent</a> as per the
+          step 6.2.2 in the <a>getUserMedia()</a> algorithm.
+        </p>
       </section>
       <section>
         <h2>
           <a>MediaTrackCapabilities</a> dictionary
         </h2>
+        <p>
+          <a>MediaTrackCapabilities</a> dictionary represents the
```
astojilj, line 619: Replace [the original sentence](http://www.w3.org/TR/mediacapture-streams/#idl-def-MediaTrackCapabilities) by:
Partial dictionary MediaTrackCapabilities extends the original MediaTrackCapabilities dictionary defined in [GETUSERMEDIA].

index.html
```diff
@@ -600,6 +629,16 @@
         <h2>
           <code>MediaTrackConstraintSet</code> dictionary
         </h2>
+        <p>
+          <a>ConstraintSet</a> dictionary specifies each member's set of
```
astojilj, line 648: <a>MediaTrackConstraintSet</a> partial dictionary extends the original <a>MediaTrackConstraintSet</a>. It specifies each member's set of allowed values.

I would keep the original sentence too here as it is sufficient to explain.

index.html
```diff
@@ -620,6 +659,10 @@
         <h2>
           <code>MediaTrackSettings</code> dictionary
         </h2>
+        <p>
+          <a>MediaTrackSettings</a> dictionary represents the <a>Settings</a>
```
astojilj, line 677: <a>MediaTrackSettings</a> partial dictionary extends the original <a>MediaTrackSettings</a> dictionary. It represents the <a>Settings</a> of a <a>MediaStreamTrack</a> object.

astojilj (commented):
The new layout looks good. Few nits.

index.html
anssiko: I prefer Oxford comma: https://en.wikipedia.org/wiki/Serial_comma :-)

index.html
anssiko: Ditto

index.html
anssiko: I think it is good to repeat the text from the main spec here for completeness. Added the proposed extra text with some modifications. Also defined "applicable constraint".

index.html
anssiko: OK.

index.html
anssiko: Added.

index.html
anssiko: Done.

index.html
anssiko: Oxford comma again :-)

index.html
anssiko: Done.

anssiko:
@astojilj, thanks for the review! I addressed your feedback in https://github.com/w3c/mediacapture-depth/pull/160/commits/5fe64304f7460975c902cd600a619028cca5e774 and will land now. We can iterate further in subsequent PRs if needed.

PR #161, opened 2017-04-05
Depth value - distance conversion. Grayscale and 16-bit depth removed.

### LINKED ISSUE #142: (16-bit) Grayscale conversion of (16-bit) depth map is wrong
With the time, the "depth map" changed the meaning from abstraction of real world distance to a 16-bit representation of depth. This part of spec remained and now is wrong - it reads as it converts 16-bit depth map to 16-bit grayscale?
Grayscale should be 8 bit, not 16-bit.

The text states:

> The **data type of a depth map is 16-bit** unsigned integer. The algorithm to convert the depth map value to grayscale, given a depth map value d, is as follows:
> 
> Let near be the the near value.
> Let far be the the far value.
> Apply the rules to convert using range linear to d to obtain quantized value **d16bit.**
> Return **d16bit.**
> 

from video:

> 6.6 The video element
> 
> When a video element is potentially playing and its assigned media provider object is a depth-only stream, the user agent must, for each pixel of the media data that is represented by a depth map, given a depth map value d, convert the depth map value to grayscale and render the returned value to the screen.
> 
> NOTE
> It is an implementation detail how the frames are rendered to the screen. For example, the implementation may use a grayscale or red representation, with either 8-bit or 16-bit precision.
>

astojilj:
Current approach is to push the original data (captured by camera)
to the WebGL textures without any conversion to a "capturable subrange"
as it would introduce performance penalty and it would not be available
on all of the devices so removing grayscale rendering.

Part of grayscale computation is also related to depth - distance:
kept and simplified.

Using 16-bit as the depth map format is not needed - normalized value
in range (0-1] is better and could handle the float point cloud to
float depth map conversion for Tango.

Fix #142

***
[Preview](https://s3.amazonaws.com/pr-preview/astojilj/mediacapture-depth/grayscale.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/mediacapture-depth/d125164...astojilj:516864a.html)


astojilj:
@anssiko , @huningxin PTAL.

Note: build.js needs to get updated before making further changes as there is a change in mathjax-node and mathjax-node-page (I've build the index.html here using older working version).

module.js:340
    throw err;
          ^
Error: Cannot find module 'mathjax-node/lib/mj-page.js'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/Users/astojilj/chromium/src/mediacapture-depth/build.js:21:13)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)

huningxin:
LGTM. Thanks for the PR!

anssiko:
LGTM2, thanks!

@astojilj, can you open an issue for build.js?

